PR #8, opened 2017-09-26
ignore timeupdates while seeking

Topsii:
I was working with high bitrate videos that take some time to load. I noticed that when seeking to some position the seeked subtitles appeared instantly while the player (videojs dash in my case) was stopped at the current frame prior seeking. Even worse the subtitles at the seeked position were playing (and would only stop after 5 seconds of not receiving update events iirc).

This fix ignores timeupdates while seeking to keep the videoimage and subtitles consistent.

### DIFF
--- a/build/subtitles-octopus/subtitles-octopus.js
+++ b/build/subtitles-octopus/subtitles-octopus.js
@@ -91,6 +91,10 @@ var SubtitlesOctopus = function (options) {
     self.setVideo = function (video) {
         self.video = video;
         if (self.video) {
+            var timeupdate = function () {
+                self.setCurrentTime(video.currentTime + self.timeOffset);
+            }
+            self.video.addEventListener("timeupdate", timeupdate, false);
             self.video.addEventListener("playing", function () {
                 self.setIsPaused(false, video.currentTime + self.timeOffset);
             }, false);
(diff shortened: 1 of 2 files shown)


Dador:
Isn't it better to just call `self.setIsPaused` in seeking/seeked event listener?
I didn't checked it yet, but looking at code it seems problem with "continue to play for 5 seconds while seeking" will remain, because you just remove sending new time to worked but don't pause the video.

Dador:
Can't confirm the issue. Please re-open with more details if you still experience the issue.

Topsii:
This commit did indeed not fix the problem.

When seeking to some position the subtitles appear immediately, even if the video-image of that position has not yet loaded. But in my opinion subtitles should only appear together with the corresponding video-image.

IIRC setIsPaused can only stop subtitles at the seek-destination, but not at the seek-origin.

As mentioned to experience this problem loading times when seeking have to be high.

What other details do you need?

Dador:
Thank you for explanation.

I thought you did a fix for issue "subtitles playing even after seeking" that you mentioned in first message, but now I understand. Anyway, sorry for misunderstanding :)

joshuabrown-ellation:
Yes please.  This would be helpful.  We could hack it by suspending the tick on animation frame on our code, but this would be a good feature add.

PR #19, opened 2017-12-01
Add conditional to update canvas style on position offset change

nickfujita:
The issue was that when the size of the video element grows and the aspect ratio is not maintained, the `top` and `left` positioning offsets of the canvas are not being updated in the `resize` function.

This PR updates the `resize` function to update the `top` and `left` position offsets when a video element is present, and also to accept these offsets as input parameters in the case that just a canvas element is provided instead.

### DIFF
--- a/js/subtitles-octopus.js
+++ b/js/subtitles-octopus.js
@@ -337,12 +337,17 @@ var SubtitlesOctopus = function (options) {
         }
     };
 
-    self.resize = function (width, height) {
+    self.resize = function (width, height, top, left) {
         var videoSize = null;
+        top = top || 0;
+        left = left || 0;
         if ((!width || !height) && self.video) {
             videoSize = self.getVideoPosition();
             width = videoSize.width * self.pixelRatio;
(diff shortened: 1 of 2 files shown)


js/subtitles-octopus.js
```diff
@@ -361,16 +373,17 @@ var SubtitlesOctopus = function (options) {
...
+                self.canvas.style.left = left + 'px';
                 self.canvas.style.pointerEvents = 'none';
             }
 
             self.worker.postMessage({
                 target: 'canvas',
                 width: self.canvas.width,
-                height: self.canvas.height
+                height: self.canvas.height,
+                top: top,
```
nickfujita: Is this sufficient for updating the canvas in the case where there is no video element provided?

Dador: No need to update top & left inside worker. Canvas inside worker is virtual.

Dador:
I've merged it (with small changes).

Thanks for contributing!

PR #29, opened 2018-10-08
Package as node module

cobarx:
Hi @TFSThiagoBR98 I made a few changes to prep for publishing on npmjs. 

I got module exports working and excluded files that won't be needed the library is consumed as a dependency. I'd suggest naming it libass-wasm since that's in line with other wasm packages. We'll want to follow npm's semantic versioning scheme for all future releases, so I figured starting with version 1.0.0 makes sense.

If you're happy with this, I have an account on npmjs already and can publish this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ instance.setCurrentTime(15); // Render subtitles at 00:15 on your canvas
 
 ### Get the Source
 
-Run `git clone --recursive https://github.com/TFSThiagoBR98/JavascriptSubtitlesOctopus.git`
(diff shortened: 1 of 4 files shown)


cobarx:
Ok we are live at:
https://www.npmjs.com/package/libass-wasm

I can add the rest of you as maintainers if you create an npm account and send me your handle.

TFSThiagoBR98:
Here my npm profile: https://www.npmjs.com/~tfsthiagobr98

PR #31, opened 2018-10-10
feat(engine): add support for changing subtitles

YePpHa:
This PR adds support for changing the subtitles without needing to re-initializing the engine. Please note that these changes haven't been applied to the sync version and it might even break it.

This PR also includes these additional changes:
- Refactor the parsing of the ass file for the font names.
- Fix issue with defining `window` in web worker scope causing issue with emscripten boilerplate code (trying to get `document.currentScript`). Not sure whether we want to include this as https://github.com/kripken/emscripten/pull/7231 will also fix the issue.

I'm not exactly sure whether the loading of fonts are working correctly. Any input on that is welcome.

This is not the exact same code as I used for [the version](https://github.com/YePpHa/JavascriptSubtitlesOctopus) I used for my [HTML5 player for Crunchyroll](https://github.com/YePpHa/crunchyroll-html5) as it also includes some "hacky" code to allow for running it in a userscript.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -196,7 +196,7 @@ src/subtitles-octopus-worker.bc: dist/libraries/lib/libass.so src/Makefile
 EMCC_COMMON_ARGS = \
 	-s TOTAL_MEMORY=134217728 \
 	-O3 \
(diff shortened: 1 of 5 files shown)


TFSThiagoBR98:
Looks interesting, i'll do some tests tomorrow

TFSThiagoBR98:
Thanks for PR

PR #48, opened 2019-06-22
Missing length function when writing fonts to file system

q-state:
Looks like it was never finding the 'Style' key because the length function was missing from the outer loop.


TFSThiagoBR98:
Thanks for PR, I'll merge it

PR #54, opened 2019-10-29
Remove '@' from font name

Teranode:
Some fonts are prefixed with @ at the start of them which makes their entire naming scheme different from the font that need's to be fetched/rendered. This will rename them to the actual font name so it's consistent.


Yay295:
Fonts that start with `@` are supposed to be displayed vertically.

Teranode:
I realize that, it still works for me regardless. I would have to re-declare the font again with the '@' prefixed to it if i didn't do this. If I don't i'd just end up with blocks

1[](https://i.imgur.com/NUkp6nM.png)

PR #62, opened 2020-01-23
Add automatic support for legacy browsers (like Smart TVs)

JustAMan:
* Change `Makefile` to produce both WebAssembly and legacy JS versions at the same time
* Add `legacyWorkerUrl` parameter to `SubtitlesOctopus()`
* Make `SubtitlesOctopus()` choose WebAssembly or legacy worker based on whether WebAssembly is supported by current browser
* Improve detection of `ImageData()` constructor (faced a case where browser lied about it being supported)
* Added a polyfill for `String.endsWith()`
* Added custom overrides for `console.log()` and friends so logging from web worker works even if the browser doesn't support it directly

All this allows me to successfully play a video with subtitles on the built-in WebOS browser starting from WebOS 1.2.x family (tested in WebOS 1.2 and WebOS 2.0 via emulators).

### DIFF
--- a/src/subtitles-octopus.js
+++ b/src/subtitles-octopus.js
@@ -103,7 +103,8 @@ var SubtitlesOctopus = function (options) {
             subUrl: self.subUrl,
             subContent: self.subContent,
             fonts: self.fonts,
-            availableFonts: self.availableFonts
+            availableFonts: self.availableFonts,
+            debug: self.debug
         });
     };
 
--- a/src/subtitles-octopus.js
+++ b/src/subtitles-octopus.js
@@ -34,6 +34,8 @@ var SubtitlesOctopus = function (options) {
 
     self.timeOffset = options.timeOffset || 0; // Time offset would be applied to currentTime from video (option)
 
+    self.hasAlphaBug = false;
+
     (function() {
         if (typeof ImageData.prototype.constructor === 'function') {
             try {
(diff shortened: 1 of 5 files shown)


JustAMan:
Note for https://github.com/Dador/JavascriptSubtitlesOctopus/issues/20 - when my PR is accepted one should be able to detect a problematic iOS version and create an instance of `SubtitlesOctopus()` with `workerUrl` pointing to the legacy (non-WebAssembly) file without the need of rebuilding things.

src/subtitles-octopus.js
```diff
@@ -1,4 +1,16 @@
 var SubtitlesOctopus = function (options) {
+    var supportsWebAssembly = false;
```
cromefire: Maybe extract that into a separate function?

src/subtitles-octopus.js
```diff
@@ -18,25 +34,36 @@ var SubtitlesOctopus = function (options) {
 
     self.timeOffset = options.timeOffset || 0; // Time offset would be applied to currentTime from video (option)
 
-    if (typeof ImageData.prototype.constructor !== 'function') {
-        (function () {
-            var canvas = document.createElement('canvas');
-            var ctx = canvas.getContext('2d');
+    self.hasAlphaBug = false;
+
+    (function() {
```
cromefire: Why is this in a self-executing function?

src/subtitles-octopus.js
```diff
@@ -108,6 +136,23 @@ var SubtitlesOctopus = function (options) {
...
+        self.bufferCanvas.height = 1;
+        var testBuf = new Uint8ClampedArray([0, 255, 0, 0]);
+        var testImage = new ImageData(testBuf, 1, 1);
+        self.bufferCanvasCtx.clearRect(0, 0, 1, 1);
+        self.ctx.clearRect(0, 0, 1, 1);
+        var prePut = self.ctx.getImageData(0, 0, 1, 1).data;
+        self.bufferCanvasCtx.putImageData(testImage, 0, 0);
+        self.ctx.drawImage(self.bufferCanvas, 0, 0);
+        var postPut = self.ctx.getImageData(0, 0, 1, 1).data;
+        self.hasAlphaBug = prePut[1] != postPut[1];
```
cromefire: Duplicate? Also again, I'd extract it.

src/subtitles-octopus.js
cromefire: Maybe extract it too

src/subtitles-octopus.js
JustAMan: I wanted to _not_ pollute the global namespace, thus all things are contained within this function.

src/subtitles-octopus.js
JustAMan: Again, this is within a nested function to not pollute the global namespace. It has single use, thus it's nameless and executed right away.

src/subtitles-octopus.js
JustAMan: Not a duplicate, I rather try to avoid a potential possibility of some code accessing `self.hasAlphaBug` before checking, so I initialize it to `false` in the constructor and compute its real value when a canvas is made

cromefire:
Well the global namespace thing is bad...

JustAMan:
@Dador @TFSThiagoBR98 any comments?

TFSThiagoBR98:
I have a question, which browsers/versions are affected by the alpha bug in WebKit?

JustAMan:
I found it in WebOS 1.2 app engine (in emulator), it runs some old-ish WebKit version.

TFSThiagoBR98:
Sorry for waiting, I'll merge this

PR #64, opened 2020-03-02
New Interface for LibASS

TFSThiagoBR98:
This PR will add support for all libass functions in Javascript side, allowing track , event and style management


JustAMan:
Hi @TFSThiagoBR98, can you provide any reasoning for adding such complexity? Why is there a need to edit events from JS side?..

PR #71, opened 2020-06-09
Simple Web Example

nickfujita:
A simple web example showing how to instantiate the sub renderer. While looking to update the to v4.0.0 in a project I'm working on from v2.1.1, it looks like it's not working as it did previously. So I created this simple web example to make sure that there wasn't something fishy with my project. However it's currently not working, and throwing a worker error:

image

Is there possibly something that has changed since v2.1.1, that no longer allows this to render on web? Tested on desktop Chrome 83.0.4103.61

Test Instructions
- In root folder start a simple server for the static assets "python -m SimpleHTTPServer 9003"
- Navigate to "http://localhost:9003/example/"
- Video should start playing, but no subtitles are visible
- Check debugger console and observe error


nickfujita:
Error was cause at mispointed subUrl. Will open separate PR for example files if decided to re-add. Converstation tracked in https://github.com/Dador/JavascriptSubtitlesOctopus/issues/70

PR #78, opened 2020-10-26
Removing the arguments.callee to support strict JS

AnonymusRaccoon:
This little change is necessary to work with scripts that have "use strict";.

### DIFF
--- a/src/subtitles-octopus.js
+++ b/src/subtitles-octopus.js
@@ -207,8 +207,8 @@ var SubtitlesOctopus = function (options) {
                 self.resize();
             }
             else {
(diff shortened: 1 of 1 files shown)


AnonymusRaccoon:
The strict mode is the default mode with many frameworks (like Angular for example). You can get information of why argument.callee has been removed in strict move [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments/callee). Can anyone merge this? Don't know if I'm pinging the right person @Dador @TFSThiagoBR98

TFSThiagoBR98:
Ok, i will merge it

PR #82, opened 2020-12-15
Initial resize fires too soon

eamanola:
Running on Web0S, that runs on a dinosaur of Chrome (ver 53.0.2785.34), on TV with no manual resize available

"loadedmetadata" fires too soon, leaving the canvas in initial display:none state.

"playing" seems like a safe place to resize


eamanola:
making a new one

PR #86, opened 2021-02-17
Add ready event upon worker-init

nickfujita:
Upon upgrading from 2.1.1 to 4.0.0, the `onReady` event appears to no longer be fired upon initialization. While on 2.1.1, it seemed that the first even was a canvas message, even before loading sub files. However, it seems that there is no such event generated on initial load, so a ready post message at the end of the `worker-init` event to trigger the `onReady` handler.


ThaUnknown:
this repo is dead and all its maintainers are completely inactive, just an FYI

TFSThiagoBR98 (approved):

Yay295:
@ThaUnknown #90

PR #91, opened 2021-05-04
Build improvements

rcombs:
- Fix non-Docker build on macOS
- Fix Docker build
- Simplify Dockerfile
- Explain non-Docker build in README
- A few other minor build-system cleanups


Dockerfile
```diff
@@ -23,7 +23,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         autoconf \
         m4 \
         gperf && \
-    pip install ply && \
+    pip3 install ply && \
```
TheOneric, line 26: The above installed packages should also be changed to their `python3-*` variants.

README.md
```diff
@@ -148,6 +148,7 @@ To use, just run: `brotli subFile.ass` and use the .br result file with the subU
 * python-html5lib
 * python-chardet
 * gperf
+* python3 ply (`pip3 install ply`)
```
TheOneric, line 151: Same `python3-*` change here.

Not required, but ideally, the dependencies would be split in deps needed only for emscripten and deps otherwise needed to build JSO. This would allow to skip superfluous manual installs when using system emscripten.

Dockerfile
TFSThiagoBR98: I think the same, @rcombs can you change to python3-ply?

TFSThiagoBR98 (approved):
I've made some changes to remove some build warnings in Harfbuzz and fontconfig

build/patches/harfbuzz/0004-disable-test-build.patch
```diff
@@ -0,0 +1,60 @@
+diff --git a/src/Makefile.am b/src/Makefile.am
+index b63bbe117..4345f0436 100644
+--- a/src/Makefile.am
++++ b/src/Makefile.am
+@@ -324,54 +324,9 @@ harfbuzz.cc: Makefile.sources
```
rcombs, line 5: Why not just run `make install-libLTLIBRARIES install-pkgconfigDATA` in the `src` dir?

build/patches/harfbuzz/0003-build-only-lib.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/Makefile.am b/Makefile.am
+index 79c271591..249da9490 100644
+--- a/Makefile.am
++++ b/Makefile.am
+@@ -4,7 +4,7 @@ NULL =
+ 
+ ACLOCAL_AMFLAGS = -I m4
+ 
+-SUBDIRS = src util test docs
++SUBDIRS = src
```
rcombs, line 10: Same

build/patches/fontconfig/0001-Disable-Tests.patch
```diff
@@ -1,13 +0,0 @@
-diff --git a/Makefile.am b/Makefile.am
-index cc3024e..e58f868 100644
---- a/Makefile.am
-+++ b/Makefile.am
-@@ -24,7 +24,7 @@
- SUBDIRS=fontconfig fc-case fc-lang src \
- 	fc-cache fc-cat fc-conflist fc-list fc-match \
- 	fc-pattern fc-query fc-scan fc-validate conf.d \
--	its po po-conf test
-+	its po po-conf
```
rcombs, line 10: Here I think you just need `install-libLTLIBRARIES` in `src` and `install-pkgconfigDATA` in the root.

rcombs (commented):
@TFSThiagoBR98 a few suggestions on your new changes; I don't think the added patch files should be needed at all.

build/patches/harfbuzz/0004-disable-test-build.patch
TFSThiagoBR98: Interesting, I didn't knew about it, i will try it out

TFSThiagoBR98:
@rcombs I made the changes you suggested, and upgraded to Emscripten 2.0.20

PR #92, opened 2021-05-08
Remove checked-in build artifacts

rcombs:
These made it more difficult to manage branches, and their purpose is better-served by github releases, the npm package, and simplified local builds.

Also adds a `prepare` script to automatically build the library (reproducibly, via Docker) on `npm pack`, `npm publish`, and local no-arg `npm install`.


package.json
```diff
@@ -7,8 +7,7 @@
     "dist/js/subtitles*"
   ],
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1",
-    "prepare": "./run-docker-build.sh"
+    "test": "echo \"Error: no test specified\" && exit 1"
```
TFSThiagoBR98: > Also adds a prepare script to automatically build the library (reproducibly, via Docker) on npm pack, npm publish, and local no-arg npm install.

@rcombs It seems that the change has been undone in this commit

rcombs: Whoops! Fixed.

PR #109, opened 2021-08-04
Bump dependecies

TheOneric:
Everything is now using the latest upstream release, except libass which is using latest master, and HarfBuzz due to problems with building newer commits. See commit msg for more details.
I briefly went to the beginning of all track+player combos from the `gh-pages`-branch with a locally built JSO to rule out obvious major regressions. 

Regardless of HarfBuzz lacking a bit behind I think overall merging this bump as is should still be useful.
I'm not familiar enough with how pthreads with emscripten is supposed to (not) work to resolve the pthreads/mutex related build problem with newer HarfBuzz-commits.


TFSThiagoBR98:
To build harfbuzz in single thread you can use this commit [here](https://github.com/TFSThiagoBR98/JavascriptSubtitlesOctopus/commit/f5e9f3a58f297b709ad5bcf3deed4867bc5f9422)

TheOneric:
> To build harfbuzz in single thread you can use this commit here

Thanks! New version, now HarfBuzz is updated to the latest release as well.

PR #15, opened 2017-10-24
Implement harfbuzz, update fontconfig, freetype, expat, and other fixes

TFSThiagoBR98:
Fixes #9 #13, and other fixes

Issue #12, opened 2017-10-20
Option to change cache size


Dador:
Add new option to change cache size.

For example right now we're using default cache size in libass (128MB + 64MB, so 192MB). Maybe it's worth to use lower default values for cache (if there is no big performance loss).

### Related links:
[ass_set_cache_limits](https://github.com/libass/libass/blob/master/libass/ass_render_api.c#L177)
[default cache size](https://github.com/libass/libass/blob/ad6510b9c78f5c79f7492b608912dc5281e15e47/libass/ass_render.h#L45)
[Relevant discussion in libass](https://github.com/libass/libass/issues/248#issuecomment-258826282)

nickfujita:
While investigating support for use in a Google Chromecast Receiver application, I think that this issue might be applicable. When running the library in conjunction with video playback, the application quickly comes to a halt, and soon after crashes the device. Looking more into the issue, I was not able to capture a memory heap snapshot while on device because the device gets locked up before the snapshot is completed. However, when running in a desktop chrome browser, I am seeing an average heap snapshot of 140MB with it's majority allocation seemingly placed on the buffer.

![image](https://user-images.githubusercontent.com/6342442/34598219-b8d4d600-f22e-11e7-9fa3-31bc1619e6a6.png)

While not sure if this is potentially a use case for the changes to cache size, and making it configurable, but thought it might be worth mentioning. If the cache size could be made configurable on initialization, it might be worth investigating, as I believe it's ideal to keep total memory usage on a Chromecast device well below 150MB total.

nickfujita:
I also tested this using only asm.js, and the results appear to be roughly the same. Would reducing the default cache size, also result in allowing us to configure a mix cache buffer size? If so, how might this affect rendering performance?

Dador:
Reducing cache limit via [ass_set_cache_limits](https://github.com/libass/libass/blob/73284b676b12b47e17af2ef1b430527299e10c17/libass/ass.h#L478) ([source](https://github.com/libass/libass/blob/master/libass/ass_render_api.c#L181)) should result in lower memory usage.
You could try it by adding call like `ass_set_cache_limits(ass_renderer, 0, 64);` in [build/subtitles-octopus/subtitles-octopus-worker.c:41](https://github.com/Dador/JavascriptSubtitlesOctopus/blob/master/build/subtitles-octopus/subtitles-octopus-worker.c#L41)
[...]

Issue #11, opened 2017-10-16
Using with flowplayer?


anoosi:
How would you use this with flowplayer?
thanks..

Dador:
Basically same as with other players: just connect it to video element.
I've added example with Flowplayer: https://github.com/Dador/JavascriptSubtitlesOctopus/blob/master/example/flowplayer.html

Issue #77, opened 2020-10-20
createEvent broken by mutating strings


ThaUnknown:
adding an event with createEvent
```js
{
Duration: 4880,
Effect: "",
Layer: 0,
MarginL: 0,
MarginR: 0,
MarginV: 0,
Name: "",
ReadOrder: 0,
Start: 7870,
Style: 1,
Text: "Text."
}
```
for some reason creates an event that looks like:
```js
{
Duration: 4880,
Effect: "Text.",
Layer: 0,
MarginL: 0,
MarginR: 0,
MarginV: 0,
Name: "Text.",
ReadOrder: 0,
Start: 7870,
Style: 1,
Text: "Text."
}
```
breaking the entire renderer, adding more subtitles with the createEvent function overwrites the text, effect and name values of the previous events

ThaUnknown:
are there any fixes for this? its becoming a serious issues for very well typesetted subs like [μtw-Kaleido] releases, which can be shown here: `60a5523d6380ed07ff63648b76324c6ca8192faf`, i currently work around this issue by building the sub file and passing it via `octopusInstance.setTrack()` but for videos with a few thousand subtitles it kills the service worker thread, even on high-end hardware!

ThaUnknown:
@Dador

TheOneric:
Can confirm this still happens with [current](https://github.com/libass/JavascriptSubtitlesOctopus/commit/9d22185464c7fb8471273609b63d8c3f29fac3dd) master.
On a side note though, you should **not** create events with a non-unique ReadOrder value. Your example sets `ReadOrder: 0` which would be valid if the track was empty before, but otherwise this has the potential to mess things up.

TheOneric:
@TFSThiagoBR98: Seems like you added this in ec21ec32597a56ff18eee4935775d32d195270f2, do you know what the "best practce"-way of passing strings from JS to WASM is? This likely affects more than just `createEvent`.

TFSThiagoBR98:
WebIDL uses this 2 generated functions:

```cpp
void EMSCRIPTEN_KEEPALIVE emscripten_bind_ASS_Event_set_Text_1(ASS_Event* self, char* arg0) {
  self->Text = arg0;
}
```

and its JS counterpart:

```js
ASS_Event.prototype['set_Text'] = ASS_Event.prototype.set_Text = /** @suppress {undefinedVars, duplicate} */function(arg0) {
  var self = this.ptr;
[...]

PR #111, opened 2021-09-26
[Jellyfin] Render ahead (oneshot render)

dmitrylyzo:
_Original author:_ @JustAMan

This PR adds a `Render ahead` mode that improves the accuracy of event timing by rendering a set of images in advance so that the frame is ready in time.
Jellyfin uses this mode.

I cherry-picked related commits, skipping merge-commits.
<details>
<summary>List of commits</summary>

848a686d8ca80dcc753ffcc8f07ebdb5591e105b
65fbfecfce503b001f55e0601c00ffa0b1f14c12
5430cc9823ca45bef41ef370eb0afef9cd078f53
4e4250204dd9b95bdc1bb3ecb67f27f587b8767e
90910bdef6e4dad06e34cfa417796583bc35cf46
21c84cc8038ea4da41652bbf22a47f18a6ead9a5
6980debc4712239f2a3c481a3a6756eca2642aca
6d3a5c731f52ea84282e40bed3014ef95220146a
7fc34cf58737992f92eb55df4d0d51b2740198e3
cea2dc9dee3d6c7607c7ec0dd48855ab59880b5a
88d2df425538467eba8d18ffc4cf1b9b3d3407f6
bd8bcab764177f5757007e606547df7397648aaa
ecd47dd25fdc7286d0bb554e99fd85c29dbc35dd
67947f395d9c6029f201e0dcc262d5b150ef1622
9dd6eb2411a18f81ad7c366224b449e6d34fb2e4
e1604cb133ef998e61e85332513301e1de008d10
`Merge pull request #6 from JustAMan/improve-timing-accuracy`
a8d670eb0955c81c6e07c99e28adac5e26c9f216
d3bc4724a0ce07f7cd02194834222999d7160a18
90d48486f1f0af0db4f1e0b17653b101781636ac
a66e79768b00b50d5a7e9651d2410677d346d688
971a9979517738ec1847350cb057d61c8840d605
a1666065e4980a7111a11817308941c4d4019560
`Merge pull request #8 from JustAMan/lite-mode`
825c9991598b33bf7f92df5c581d39d427cb6cba
345d7010b1d31f933f1b2cc3a17d83a09b5e5c2e
4c5f018a07d5c51b089ef21ccf1f81a0355e6c8c
bcf4b5fc62140a3ab37ce5e3f6c329c3094cb7d8
5256b6364941f7f77b547493141d59c9efc89d0d
`Merge pull request #11 from JustAMan/limit-canvas-size`
cc9575cd5137c97c347940bc7f9d54822358c133
93f4d0f4a77f60133a6db4b60dc7f135b7218daf
0722559b28ea3e9e1eeedc28b7f751439f345563
`Merge pull request #12 from JustAMan/fix-resize`
7eefe74ca28fd4fd46e1d9db475f2dc7f110aa42
595d0e20e58634130c237a2853be70fa193c530d
`Merge pull request #15 from JustAMan/update-build`
c66e7b1841118f957301e5b950d31c5e9aa32b6e
`Merge pull request #19 from Teranode/master`
0e9e7929dddbbf6168f43032ff90f6b86fd5c318
9eac714a26d6d57b83ed1274d7a45da160786710
</details>

**Changes**
Add `Render ahead` mode.


ThaUnknown:
@dmitrylyzo could you resolve conflicts so we can get this moving?

TheOneric:
Thanks again for working on this! Here are some remarks after just a first cursory look:

 - Please indicate the original commits this was cherry-picked from in the commit-message as done in #112.
 - Needs to be rebased
 - This seems to do much more than just "prerendering frames for later deployment"; it isn't immediately clear in all cases if those other changes (limiting ca
nvas size, dropping animations, ...) are part of the oneshot render mode, or when they are active
 - Can you clarify if the new rendering modes are also droppng frames (like `lossyRender`) or blending in WASM (like `blendRender`) instead of JS?
 - All those modes and new options need to be documented in the README, but none currently are
   *(tbf, the `blendRender` mode also wasn't documented until recently, but now it is: 9424e8f83359e7aa1881336800df61a6580be755)*
 - At first glance this seens t limit the canvas size *by default*; I'm not sure we should do that *by default*.

dmitrylyzo:
> This seems to do much more than just "prerendering frames for later deployment"; it isn't immediately clear in all cases if those other changes (limiting canvas size, dropping animations, ...) are part of the oneshot render mode, or when they are active

iirc, it was tested on heavy animated subtitles and these features were born in between. They are too dependent on the oneshot render code base.
You can see the `List of commits` in the first post. I marked merge-commits in our fork.

> Can you clarify if the new rendering modes are also droppng frames (like lossyRender) or blending in WASM (like blendRender) instead of JS?

This is a single mode and it uses `blendRender`. https://github.com/libass/JavascriptSubtitlesOctopus/blob/61c8997337a2d51b4117a5e0cc85e75ff48348af/src/post-worker.js#L260
It can erase animation tags, but I haven't tested that. Jellyfin doesn't use it. So it was probably only tested in the dev cycle.
<details>
<summary>Current Jellyfin options</summary>

https://github.com/jellyfin/jellyfin-web/blob/f4536f4865a6d3110db885159a95befbf0f14dfe/src/plugins/htmlVideoPlayer/plugin.js#L1075-L1085
```
                // new octopus options; override all, even defaults
                renderMode: 'blend',
                dropAllAnimations: false,
                libassMemoryLimit: 40,
                libassGlyphLimit: 40,
                targetFps: 24,
                prescaleTradeoff: 0.8,
                softHeightLimit: 1080,
                hardHeightLimit: 2160,
                resizeVariation: 0.2,
                renderAhead: 90
```
</details>

> At first glance this seens t limit the canvas size by default; I'm not sure we should do that by default.

This?
https://github.com/libass/JavascriptSubtitlesOctopus/blob/61c8997337a2d51b4117a5e0cc85e75ff48348af/src/subtitles-octopus.js#L23-L25
if `prescaleTradeoff` is `null` (by default), `hardHeightLimit` is only used.
https://github.com/libass/JavascriptSubtitlesOctopus/blob/61c8997337a2d51b4117a5e0cc85e75ff48348af/src/subtitles-octopus.js#L705-L708
This is probably done to avoid wasting time rendering too high quality subtitles.

ThaUnknown:
I've been busy, so I wasn't able to leave feedback on this.

Is there any reason to force the use of blend render? It performs a lot worse than fast render since it can't benefit from hardware acceleration. I think it should inherit the rendering mode which was user specified on init, rather than force one.

Also from my testing SO performed worse with pre-rendering on releases that had >300 bitmaps per frame, but I'm not too sure if I made some sort of mistake, or if it simply can't handle it.

Shouldn't all the pre-render logic be handled on the worker? Doing it on the main thread seems very wasteful.

IMO, the default hard size cap should be set to 2k, for retina displays.

dmitrylyzo:
> renderAhead in README claims to be directly related to frame count; but the commment in subtitles-octopus.js says it relates to how many MiB's of data get (pre-rendered and) chached for later use. Which one is correct?

From the option list in README:
- `renderAhead`: How many MiB (approximate) of subtitles to render ahead and store.

> The added description of renderAhead reads to me like it is supposed to just prerender all timespans which ahve any event; but the animation bugs you uncovered earlier — which iiuc appeared when using the renderAhead mode — and a glance at libassjs_find_event_stop_times suggest it also tries to be smart regarding animated events. Can you clarify what it actually does (and amend the documentioan commit accordingly)?

_Hmm, `timespan` is a more correct term._
It builds a set of timespans where nothing changes. _As I understand._
```
Dialogue: 0,0:00:00.00,0:00:05.00,Default,,0000,0000,0000,,1
Dialogue: 0,0:00:03.00,0:00:08.00,Default,,0000,0000,0000,,2
```
There will be 3 timespans (by design): `1`, `1+2`, `2`.
In fact, there are more:
```
0: 0..3                       1
1: 2.999..3.0323333333333333  
2: 3.0323333333333333..5      1+2
3: 4.999..5.032333333333333   
4: 5.032333333333333..8       2
```
The timespans in between are because of this time shifting (_not sure of its purpose_): https://github.com/libass/JavascriptSubtitlesOctopus/blob/58306888a286d8e9680e6290f6fc99d5cb7ceb56/src/subtitles-octopus.js#L330
Next, it needs multiple timespans for animation. Because of that bug, the oneshot renderer doesn't know it should make them, and only does the first one.

> and in fact I don't know why ; are treated special at all, afaik they only have a special meaning inside the Effect field not in override-tags

I just left it as it was used. But I didn't see it in the documentation.

> I suggest to only check for the backslash followed by the name of an override tag and not any of the subsequent cahracters.

Well, if only to expect correctly written tag names. Or maybe I can find the end of the word :thinking:
`{\t (\frz360}` - doesn't the missed `)` bracket make sense?

TheOneric:
> It builds a set of timespans where nothing changes. *As I understand.*
> […]
> Next, it needs multiple timespans for animation. Because of that bug, the oneshot renderer doesn't know it should make them, and only does the first one.

So yes, it does try to be clever *(and partially fails)*. Putting aside if that even makes sense to begin with (libass also caches parts of events, more for non-animated then for animated), how about replacing the second-last sentence in `#### Render Ahead` with:
```
This mode will analyse the events to avoid rendering empty sections or rerendering non-animated events.
```
> Well, if only to expect correctly written tag names. Or maybe I can find the end of the word :thinking:

The examples I gave *are* correct ASS. I think just not checking following characters is the safest, **if** you really want to I think checking for *either* `(` or ` ` to follow *maybe* works for complex tags (which *require* multiple arguments), but I'm not entirely sure about that and it definitely does **not** work for non-complex tags. *(So again, not checking that is safest and false-positives should be sufficiently rare.)*

> `{\t (\frz360}` - doesn't the missed ) bracket make sense?

The example is correct, if it extends to the end of the override block it doesn't need a closing parenthesis `)`.

TheOneric:
Looking at the revised “Fix event tag parsing” I can again see more parsing mistakes in the newly touched code *(eg using locale-dependent functions, assuming parentheses can be nested, and more)*. Rather than analysing and fixing all of this I rewrote the tag parser, please replace the “Fix event tag parsing” commit with this one: [fix-animation-detection.patch](https://github.com/libass/JavascriptSubtitlesOctopus/files/7332058/0001-renderAhead-dropAllAnimations-fix-animation-detectio.patch.txt) *(apply on top of 8159a6b with `git am <patch-file>`)*.
I tested it on this file: [animated.ass](https://github.com/libass/JavascriptSubtitlesOctopus/files/7332071/animated.ass.txt) and for reference here's the expected and produced output with `0` for unanimated, `1` for animated` and `clear` shows what dropping the animations does:
```
1:     Move (with tabulator) {\move (50,50,1230,670)}
clear: Move (with tabulator) {                      }
0:     {\c1&H000000&\alpha&HAA}Move (with unclosed block) {\move(50,50,1230,670)
1:     {\notatag \fs50 \t     (\frz360}Transform (with space)
clear: {\notatag \fs50         \frz360}Transform (with space)
1:     {\frz30 aa \fad(100,500) \fs120}Fad simple
clear: {\frz30 aa               \fs120}Fad simple
1:     {\frz-30 b \fade(50,255,0,10,300,700,900) \fs121}Fade complex
clear: {\frz-30 b                                \fs121}Fade complex
1:     いろは{\kf128}笑{\kf47}え{\kf24}ば {\kf91}空{\kf58}ま{\kf27}で
clear: いろは{      }笑{     }え{     }ば {     }空{     }ま{     }で
1:     にほえと{\ko128}笑{\ko47}え{\ko24}ば {\ko91}空{\ko58}ま{\ko27}で
clear: にほえと{      }笑{     }え{     }ば {     }空{     }ま{     }で
1:     ちりぬるを{\K128}笑{\k47}え{\K24}ば {\k91}空{\K58}ま{\k27}で
clear: ちりぬるを{     }笑{    }え{    }ば {    }空{    }ま{    }で
0:     {\shad5\bord5\fnIPAMincho}週間
0:     {\an2}Not animated due to libass ext \{\move(1,1,500,500)}
1:     {\an8}{\bord3}Animated {\move(1,1,500,500)} with unusual{\1c&H00FF00&}ly placed {\r}tags
clear: {\an8}{\bord3}Animated {                  } with unusual{\1c&H00FF00&}ly placed {\r}tags
```

*(As a side note, assuming emscripten could utilise SIMD in its library function it would probably be more performant to adjust `_is_event_aniamted` and `_is_block_animated` to use [`strchr`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/strchr.html) rather than a for-loop over all chars. But last I heard, WASM-SIMD is still not part of the standard but a browser-specific extension, so this probably doesn't matter — and I'm not sure if the for-vs-SIMD performance difference would even be noticeable here to begin with)*

TheOneric:
> But I tested with the closed \t:
> [it includes a] ^ closing parentheses because \frz360 skips further parsing of \t.

The closing parentheses doesn't matter as it will just be ignored in parsing.

Here's an updated version of the commit with a more correct `length` in `_is_animated_tag` and fixing a misplaced `block_start = NULL;` in `_is_event_animated`. Can you keep the commit-title and message intact? *(If you use `git am` it will do this by default)*: [fix-animation-detection.patch.txt](https://github.com/libass/JavascriptSubtitlesOctopus/files/7358039/fix-animation-detection.patch.txt)

Can you also add an explaining description to the commit messages of `Fix a lost last event` and `Fix blend render returning empty result`? That is: what was the problem, which render-mode and/or settings did it affect and why is this a good fix. This will make things easier when looking back at those commits in the future when the context is already forgotten.

TheOneric:
> > Can you also add an explaining description to the commit messages of `Fix a lost last event` and `Fix blend render returning empty result`? […]
> They are simple checks.

With the context in mind they are obvious yes. But when I looked at them again today, a week later, I already forgot about the context and first had to look a bit at the surrounding lines to again understand why this particular early return with all empty images fixes “blend render returning empty result”, or in fact why an empty result is a problem. Putting two explanatory sentences in the commit messages is quick and saves time later on if someone looks at it again in the future.
`Fix a lost last event` is simpler, but when in doubt it is imho better to err on the side of additional explanation rather than potentially being stuck without needed/helpful explanation/context in the future.

> Would you ask about them if they were part of the original PR?

If you mean the backported patches, then in the current situation I'm not asking for those to be changed because they've been deployed in jellyfin for 1+ year, so there's some worth in preserving the actual commit history close to as was.
If however new patches were posted for review/merge like this and I were to decide, I would most definitely insist on the commits being amended to achieve *well documented* and atomic commits first.

> FYI, if you are going to do a "rebase and merge" (for linear history), there are commits that are not atomic.

Yeah, already noticed but thanks for pointing this out again. I'll try to remember to use a merge commit. :)

dmitrylyzo:
`Bad blend result` can be reproduced (without a corresponding commit) with this [test5-empty.ass.txt](https://github.com/libass/JavascriptSubtitlesOctopus/files/7359607/test5-empty.ass.txt) and `renderMode: 'blend', dropAllAnimations: true, renderAhead: 0`.
It might be better to fix in libass, so that it returns `NULL` (blank frame) and doesn't return empty sub-images.

_Emscripten test is broken._ :confused:
_Maybe rerun?_

TheOneric:
@dmitrylyzo : I just tested the AoT Karaoke in Firefox-ESR 91 with
```
    renderMode: 'blend', // also without
    renderAhead: 9, // also 1 and 90
```
The subtitles only update when I pause and unpause the video. Without `renderAhead` everything works as expected.

dmitrylyzo:
Unlike `fast` (lossy), render ahead doesn't skip frames. That's why so animation intensive subtitles such as AoT may stuck.
If worker unable to handle it, you see `worker busy, requesting to seek` in the browser console. In this case `renderAhead` is irrelevant - there will be no ready frames.

If you pause and wait, you will see some `oneshot received (...), render: XX ms` - filling the pre-render cache. When you unpause, you should see updates until the pre-render cache is empty.

Screen resolution plays a lot.
At 1920x1080 `Toaru Kagaku no Railgun S` `renderAhead: 90`, I see no problem, because render time fits FPS.
```
oneshot received (start=16.30820666666668, empty=16.31), render: 15 ms
oneshot received (start=16.341540000000016, empty=16.35), render: 7 ms
oneshot received (start=16.37487333333335, empty=16.38), render: 7 ms
oneshot received (start=16.408206666666686, empty=16.41), render: 4 ms
```
But at 3840x2160 it stuck.
```
oneshot received (start=13.305811, empty=13.36), render: 157 ms
(4) worker busy, requesting to seek
oneshot received (start=13.461356, empty=13.56), render: 68 ms
(3) worker busy, requesting to seek
oneshot received (start=13.534487, empty=13.56), render: 56 ms
```

TheOneric:
> Screen resolution plays a lot.
> At 1920x1080 Toaru Kagaku no Railgun S renderAhead: 90, I see no problem

I didn't fullscreen it, so the physical size should have been below 1920×1080.

> Unlike fast (lossy), render ahead doesn't skip frames. That's why so animation intensive subtitles such as AoT may stuck.

Lossy renderer behaves similarly for me, just not quite as bad (this hasn't chagned) and the doc for the renderer already mentions that it will drop frames if they aren't ready in time.
Does `renderAhead` try to fill the *whole* prerendering cache before it actually shows anything? If so why? Intuitively I'd expect the prerender cache to fill up independently from the current rendering; if the current frame hasn't been prerendered yet, just render it *now* *(and move the prerendering a bit into the future, possibly using the current desynchronisation as an indicator for how far to skip ahead)* — this way it shouldn't be *worse* than plain `blend`.

dmitrylyzo:
> Lossy renderer behaves similarly for me, just not quite as bad (this hasn't chagned) and the doc for the renderer already mentions that it will drop frames if they aren't ready in time. Plain blend is sometimes missing a few frames on heavy sections but will catch up again and behaves the best so far.

AFAIK, the problem with the other modes is that they start rendering at a time when the frame should already be displayed.
If we render fast enough, we don't notice the lack of sync, and `renderAhead` isn't needed.

> Does renderAhead try to fill the whole prerendering cache before it actually shows anything? If so why? Intuitively I'd expect the prerender cache to fill up independently from the current rendering;

When the video is ready, oneshotRender starts requesting the worker for timespans (`oneshot-render`), and the worker responds with a ready frame for the specified timespan (`oneshot-result`). This process runs in parallel until the `renderAhead` limit is reached. But if the worker renders too long, the pre-render cache stops filling, and we get stuck.

> if the current frame hasn't been prerendered yet, just render it now (and move the prerendering a bit into the future, possibly using the current desynchronisation as an indicator for how far to skip ahead) — this way it shouldn't be worse than plain blend. Is there already something like this planned?

For on-demand rendering we need a second worker - the current one is busy.
@JustAMan was thinking about multithreaded rendering, iirc. But no one has tried to do that since then.

TheOneric:
Just to be sure I tried again with controlled physical canvas sizes, and I noticed that the freezing only happens if the console is already opened *While* loading the page. If the console is opened later or never it renders smoothly. Any thoughts about that?

dmitrylyzo:
> Just to be sure I tried again with controlled physical canvas sizes, and I noticed that the freezing only happens if the console is already opened _While_ loading the page. If the console is opened later or never it renders smoothly. Any thoughts about that?

I can correlate the smoother behavior with the closed console (no debug printing), but with the moment it is opened :man_shrugging: 

> Thanks for the explanation, but I'm afraid without reading the code I'm still not sure about why/how the freeze can come to be. Do all frames of the requested timespan or the cache need to be finished before something is displayed?

It requests them one by one.
This function is called each frame, and if there is a ready frame in the cache, it displays it.
https://github.com/libass/JavascriptSubtitlesOctopus/blob/c8e560cc412d360c042ebf5ec36d36aea230a592/src/subtitles-octopus.js#L375-L408
Then, after the response, `tryRequestOneshot` is called to request the next timespan.
https://github.com/libass/JavascriptSubtitlesOctopus/blob/c8e560cc412d360c042ebf5ec36d36aea230a592/src/subtitles-octopus.js#L303-L346

> When it get's delayed how is the prerendering adjusted?

I am not sure here.
There is a `requestNextTimestamp` property that is used to move on if we get stuck. But there is no adjustment for the delay.

TheOneric:
Yeah, just displaying the last already rendered event may be a simpler fallback. Together with moving the prerendering start to the next (or second or third next) frame (instead of the current frame) and maayyybe also lowering the framereate of animations until the cache catches up to `currentTime` again, this should *in theory* prevent freezes and not be worse than plain `blend`. *(Eg half the current framerate  — but at least 1fps — each time a non-prepared frame is requested; when prerendering caught up again reset the framerate to it's original value)*

Yay295:
Just a note on parsing transitions: the start of a new transition ends the current transition. I'm not sure if that's already handled. For example: `\t(0,130,\1c&HFFFFFF&\t(65,565,\fscy0)` is two transitions.

ThaUnknown:
why not 23.976? it ends up being a floating point anyways, so there isn't any perf lost, and it's a bit more accurate to the video framerate

TheOneric:
> why not 23.976? it ends up being a floating point anyways, so there isn't any perf lost, and it's a bit more accurate to the video framerate

Only if it's an NTSC-video. For PAL and SÈCAM *(and the in recent years also appearing exact `24fps` videos)* it will be less in-line with the video framerate. Not that I think a difference of <2fps matters for browser-rendering, where we can't perfectly sync subs and video anyway.

dmitrylyzo:
> perhaps also give the prerendering chache more structure than a flat map of frames?

Sorry, didn't get about the cache structure. :sweat_smile: 

Currently, it is already an ordered array of object with such info:
```js
{
    eventStart: <number>, // start of the frame (show)
    eventFinish: <number>, // end of the frame (hide)
    emptyFinish: <number>, // end of the gap after
    spentTime: <number>, // time spent in total (for debugging)
    blendTime: <number>, // time spent by blending (for debugging)
    viewport: {
        width: <number>, // width of the frame
        height: <number> // height of the frame
    },
    items: <array of images>, // items to display
    animated: <boolean>, // this frame belongs to animated event (and needs to be subdivided)
    size: <number> // size in bytes
}
```

TheOneric:
> Currently, it is already an ordered array of objects with such info:

Oups, I mistakenly thought those objects were always single frames, but `items` is already an array. Nevermind then.

ThaUnknown:
> As this adds multiple options and changes&adds a lot of code, it would be great if I could get another opinion from a JSO-maintainer on this. A summary can be found below.
> 
> As a note for testing: Opening the browser-console makes rendering much slower due to the debug prints, which may trigger the (documented and planned to be improved) subtitle-freezes with `renderAhead`.
> 
> The bulk of the changes are backported jellyfin-commits, those add three separate things: 1.) `renderAhead` to prerender 2.) An option to remove all animating tags 3.) Options to bound the size (in pixels) of the rendering canvas _(which will then be scaled by the browser to fit the physical size)_
> 
> To my understanding it is difficult to split these up because the original commits are interspersed and depend on each other and the commit messages are not always accurately describing what it actually contains and they are definitely not always atomic. _(but there's some worth in keeping close to the original commits as they are used in the jellyfin-fork for 1+ year)_. All of this makes them not only hard to split up but also a major pain to review. The backports are followed up by new commits fixing various issues found in testing and glancing over the code. It appears to pass all our samples from the gh-pages branch.
> 
> .
> 
> Future todos, probably fine for **after** this has been merged (in no particular order):
> 
> * Add a `renderAhead` sample to our example suit _(possibly duplicate the AoT one so it's possible to compare `renderAhead`+`blend` with `lossy`)_
> * investigate whether we can make the WASM-blendRenderer the new default renderer; it seems as it should always be faster than the current default and afaik there are no frame-dropping or browser-incompatibility concerns with it unlike with the lossy "`fast`"-mode or `renderAhead`
> * improve `renderAhead`'s behaviour when falling behind the video _(the goal is to not be _worse_ than using a plain blend redererer)_
> * check if `renderAhead` can be made to work with other rendering-modes
> * tiled blending patches from jellyfin
> * DataHoarder's font loading improvements
> * low-prio: simplify and improve the resizing logic for the prerendering cache
>   
>   * after that: maybe consolidate the (remaining) `renderAhead` related options into one obejct

- WASM blend render isn't worth it on high end hardware, it can't benefit from hardware accelerated canvas blending, so while it might offer insane results on low end hardware, it probably wont on high-end hardware like mine, I tested this on subtitles which  had ~600-1k bitmaps per frame, which performed... okay on my threaded render, but almost not working on blend render
- a bit off-topic, but with how stuff is currently going in web, streaming should be a high priority, so we should probably investigate how this behaves on adding events, speaking of which, completely aren't functional https://github.com/libass/JavascriptSubtitlesOctopus/issues/77 and if I remember correctly, renderAhead didn't work well if events were added after it thought there are no more events to pre-render

TheOneric:
> > > > investigate whether we can make the WASM-blendRenderer the new default renderer; it seems as it should always be faster than the current default
> > > 
> > > 
> > > WASM blend render isn't worth it on high end hardware, it can't benefit from hardware accelerated canvas blending, so while it might offer insane results on low end hardware, it probably wont on high-end hardware like mine […]
> > 
> > 
> > Defaults are supposed to work well _everywhere_, not just on some special hardware+software combination.
> 
> yes, that is what I'm saying, WASM blend won't work for high intensity TS

Maybe I should've been more clear: Defaults should work *reasonably* well on *all devices and engines*. WASM-Blending does and so far seems to always be better then the current default. `lossy` meanwhile has serious problems on many devices and engines.

ThaUnknown:
> > > > > investigate whether we can make the WASM-blendRenderer the new default renderer; it seems as it should always be faster than the current default
> > > > 
> > > > 
> > > > WASM blend render isn't worth it on high end hardware, it can't benefit from hardware accelerated canvas blending, so while it might offer insane results on low end hardware, it probably wont on high-end hardware like mine […]
> > > 
> > > 
> > > Defaults are supposed to work well _everywhere_, not just on some special hardware+software combination.
> > 
> > 
> > yes, that is what I'm saying, WASM blend won't work for high intensity TS
> 
> Maybe I should've been more clear: Defaults should work _reasonably_ well on _all devices and engines_. WASM-Blending does and so far seems to always be better then the current default. `fast` meanwhile has serious problems on many devices and engines dropping frames and in some cases even being slower than the current default.

lossy performs well on 70% of cases, according to 2021's browser market shares [chrome, edge, firefox], from my testing WASM blending had much more mixed results than fast or even normal, I'd have to re-run these tests after my performance improvements vs this new blend, because the current normal or fast performance is a joke

rcombs:
This PR remains too large to review practically in a single shot. Furthermore, some of the later commits are fixes for bugs introduced in previous commits (e.g. 10c35c587d10ed593b647f1e2cd144fe3736cd59 appears to fix a bug introduced in or after 95bff90e2e46ec3bbd3cf0ddd940c9cf85ddfa30), meaning that the repo state is broken between commits, which makes this unsuitable for merge as-is. Never break functionality in one commit and then fix it in another within the same PR; just rebase and fix the broken commit.

Some changes from this are not actually relevant to this branch, and can be broken up into smaller PRs for separate review and merge review; for instance:
- All style cleanup (whitespace changes, etc) is separable from actual code changes
- Changes to default settings (e.g. frame rate 30->24) are separable from everything else

Furthermore, some of these commits refactor code that was added in previous commits within the same PR. This makes commit-by-commit review effectively useless. For instance, f38fc7e0ddc37e63b337a0f79b3ddd073bacfabf deletes large amounts of code added in 69906fcc30cf28678cc428db2801f23b44cdffc1. Please squash your commits down appropriately.

Meanwhile, many of these commits add both base-level infrastructure and calls into it in the same commit. This makes review more difficult, and also would make future bisects harder. Implement new C++ functionality in one commit, then call it from JS in another, with each split up into as many commits as appropriate (for instance, adding a new class should usually be a commit on its own).

dmitrylyzo:
Most of the commits are not mine - I am just porting a feature from another repo and trying to keep the original authorship of the code.

FPS change is added after https://github.com/libass/JavascriptSubtitlesOctopus/pull/111#discussion_r731431294

ThaUnknown:
> Most of the commits are not mine - I am just porting a feature from another repo and trying to keep the original authorship of the code.
> 
> FPS change was added after [#111 (comment)](https://github.com/libass/JavascriptSubtitlesOctopus/pull/111#discussion_r731431294)

I think the original commit structure doesn't matter, just what what does, as long as you put justaman as a co-author it should be fine, but from what I talked about with rcombs, a feature/functionality shouldn't be split into more than 1 commit, or more simply the same code can't be changed in more than 1 commit, and unfortunately this won't be merged until those requirements are met ;-;

so if I was you I wouldn't worry that much about the original commit structure, just focus on the functionality

JustAMan:
I'm fine with any editing of the commits, I can recede the "code ownership/authorship" to anyone wanting to maintain this if this helps going further.

I did this in the first place "for the greater good", not for the fame or anything.

TheOneric:
> Most of the commits are not mine - I am just porting a feature from another repo and trying to keep the original authorship of the code.

There's some value in keeping a close mapping to the original history since it was already deployed publicly for over a year; that's why this form wasn't rejected immediately. However, the non-atomicity and interleaving of multiple changes is too detrimental to the maintainability and bisectability in the long term, while in the short term it makes it very hard to properly review (as the long response times show), making bugs slipping in more likely *(we already identified & fixed several bugs, possibly there are even more)* and hampering a merge.
Thus, it was decided to forgo preservation of the original history and insist on the general requirements for patches also here.
As for the matter of authorship, there's the widely recognised `Co-Authored-By: ` commit message trailer which can be used to indicate that multiple people worked on a change, with the original or largest contributor for a particular commit being granted the primary authorship. *(I just noticed, you already used this in one commit.)*

.

I think, now that we're no longer trying to emulated the original jellyfin-history, this can be split into at least four series (in whatever order is most convenient):

 - `renderAhead` mode; possibly also including the proposed changes to its behaviour when falling behind ( + a reminder or patch to add it to the sample suite)
 - the option to bound the canvas size
 - the option to drop animations
 - change `fps` default, make the WASM-blendRenderer the new default mode and other default changes if there were any
 
 Thanks again for your work on this JustAMan and dmitrylyzo!
 
 ---
 
 Just so it isn't buried under all the comments posted since, here are [again](https://github.com/libass/JavascriptSubtitlesOctopus/pull/111#issuecomment-948920513) the future todos previously identified. The first three *(as noted above)* and the consolidation of `renderAhead`-options would probably be best integrated into the newly split series. I don't recall otoh what the resize logic improvements were about, if needed they can be another series before the `renderAhead` one, to avoid introducing option that change shortly after.
 
> Future todos, [...] (in no particular order):
> 
>   * Add a `renderAhead` sample to our example suit _(possibly duplicate the AoT one so it's possible to compare `renderAhead`+`blend` with `lossy`)_
> 
>   * investigate whether we can make the WASM-blendRenderer the new default renderer; it seems as it should always be faster than the current default and afaik there are no frame-dropping or browser-incompatibility concerns with it unlike with the lossy "`fast`"-mode or `renderAhead`
> 
>   * improve `renderAhead`'s behaviour when falling behind the video _(the goal is to not be _worse_ than using a plain blend redererer)_
> 
>   * check if `renderAhead` can be made to work with other rendering-modes
> 
>   * tiled blending patches from jellyfin
> 
>   * DataHoarder's font loading improvements
> 
>   * low-prio: simplify and improve the resizing logic for the prerendering cache
>       
>     * after that: maybe consolidate the (remaining) `renderAhead` related options into one obejct

dmitrylyzo:
After squashing, you lose all messages of intermediate commits and probably info about `why is that`. _just to warn_

Bisecting: there is probably not much difference between a single (squashed) commit and a merge commit when using the `--first-parent` option (git 2.29).

The `Co-Authored-By:` doesn't specify the exact part of the code that the co-author touched. _I need to know who to blame (even if it's me)._
The one that exists is because I reviewed the PR.

`dropAnimation` feature is a bonus to the animation detector (parser).

Since I will port using the same`original, then fixes` pattern, I think you won't accept the PR with the canvas bounding feature as well.
https://github.com/libass/JavascriptSubtitlesOctopus/pull/111/commits/f2c1fead9ba3dea16a2e2b613556fe33e266a6f4
https://github.com/libass/JavascriptSubtitlesOctopus/pull/111/commits/35efbb9b557d63247097bb06627bc16e3735cfb4
https://github.com/libass/JavascriptSubtitlesOctopus/pull/111/commits/a602fd7f68602e1c59a04609f4c93463dd3727c1
https://github.com/libass/JavascriptSubtitlesOctopus/pull/111/commits/c2b711a3f4756a5e789d59372e8e15d6329bfeeb
part of https://github.com/libass/JavascriptSubtitlesOctopus/pull/111/commits/40da83b7063202b247651db5fc940b47093ec1c8

I think you need someone with a more flexible mind.

TheOneric:
> `dropAnimation` feature is a bonus to the animation detector (parser).

Adding the animation detection and the option first and then later reusing the detection for `renderAhead` brings the same end result, but as a bonus takes some bulk away from the `renderAhead` series.

> Bisecting: there is probably not much difference between a single (squashed) commit and a merge commit when using the `--first-parent` option (git 2.29).

There is a significant difference in all but single-commit merges.
With `--first-parent` the whole merge is effectively a single (enormous) commit, if the commits making up the merged series are not working properly individually, the series itself cannot be bisected.

A proper patch series consists of well-documented atomic commits, meaning:
 - each commit compiles individually
 - each commit is, at the time of merge, assumed to be correct by itself
 - the state of the whole repo makes sense after each individual commit
 - there are no *temporary* user-visible behaviour changes within a series
 - each commit is as concise/small as possible without breaking any other requirements
   (among other that means, refactors are split from the actual code changes)
 - each commit with a non-trivial change explains in its commit message what it does, why and what assumptions and/or considerations went into the specific implementation of the change

This not only makes review easier, it also means the patch-series itself can be bisected and when you get a result, the amount of code needed to analysse for the source of the bug is much smaller and you have plenty of explanation for what it wants to do and why it was done the way it is, helping in understanding the code and making it less likely new bugs are introduced in an attempt to "fix" the bug currently at hand.
Using `--first-parent` enables none of this.

Sometimes this can mean the commit message is longer than the actual code change like [here](https://github.com/libass/libass/commit/d7e0e17ce571a8eca8fdf978a7c00c096782d037). For an example of how the commits are supposed to be split up (but all commits are almost trivial) see the `renderMode` option unification I split out of this series as an example: #118.

> The Co-Authored-By: doesn't specify the exact part of the code that the co-author touched.

If relevant this info can be added to the commit message.

> *I need to know who to blame (even if it's me).*

Pointing fingers to blame when a bug is found is not productive. Fixing the bug is.
Properly documented atomic commits make it easy to locate regressions and to understand how they came about.
An amalgamation of potentially non-compiling and individually bugged commits, makes all of identifying, understanding and fixing harder. Willingly making the code harder to maintain just to make finger-pointing easier is not merely unproductive but detrimental.

PR #113, opened 2021-09-26
Fix blend render after seek

dmitrylyzo:
Blend render doesn't redraw the frame if you seek on an "empty" frame - the previous image remains on the screen.

**Steps To Reproduce**
1. Use `blend` render mode (`blendRender: true`).
2. Start a video with subtitles that doesn't start at the beginning.
3. Wait for the subtitles to appear.
3. Seek on an "empty" frame (at the beginning or any other position without subtitles).
4. The subtitles don't change to blank.

### DIFF
--- a/src/SubtitleOctopus.cpp
+++ b/src/SubtitleOctopus.cpp
@@ -233,10 +233,11 @@ public:
 
     RenderBlendResult* renderBlend(double tm, int force) {
         m_blendResult.blend_time = 0.0;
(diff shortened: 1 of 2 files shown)


TheOneric:
Thanks, can confirm the issue and that this fixes it. The changes seems innocuous enough to me, but let's wait a while if Thiago or rcombs want to have a closer look at it.

TFSThiagoBR98 (approved):

PR #114, opened 2021-09-30
Legacy Brotli

dmitrylyzo:
1. A bunch of polyfills need to be added to support Brotli compression in legacy browsers.
Some of them (marked `From brotli`) are from https://github.com/google/brotli/blob/master/js/polyfill.js So they can be added by this file passed in `emcc`.
At the same time, if something changes in Brotli polyfills, it may affect JSO. For example, there is no `Int8Array.slice` in the version currently used by JSO, but there is in the latest recent.
Therefore, it is probably better to rely on Emscripten polyfills and explicitly added JSO polyfills.

2. URLs can contain params. For example, in Jellyfin the URL looks like `/Videos/{1}/{2]/Subtitles/{3}/{4}/Stream.ass?api_key={5}`.
_Jellyfin doesn't support Brotli compressed subtitles yet, but maybe someday._
_If you want, I can extract this part in a separate PR._

Tested in webOS 1.2 emulator with Brotli subtitles from `gh-pages` branch.

**Some thoughts**
Maybe extract polyfills in `polyfills.js` and add them only to the legacy worker?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,6 +70,7 @@ $(DIST_DIR)/lib/libexpat.a: build/lib/expat/configured
 	emmake make install
 
 build/lib/brotli/js/decode.js: build/lib/brotli/configured
+build/lib/brotli/js/polyfill.js: build/lib/brotli/configured
 build/lib/brotli/configured: lib/brotli $(wildcard $(BASE_DIR)build/patches/brotli/*.patch)
 	rm -rf build/lib/brotli
 	cp -r lib/brotli build/lib/brotli
@@ -325,10 +326,11 @@ dist/js/subtitles-octopus-worker.js: src/subtitles-octopus-worker.bc src/pre-wor
 		-s WASM=1 \
 		$(EMCC_COMMON_ARGS)
 
-dist/js/subtitles-octopus-worker-legacy.js: src/subtitles-octopus-worker.bc src/polyfill.js src/pre-worker.js src/SubOctpInterface.js src/post-worker.js build/lib/brotli/js/decode.js
+dist/js/subtitles-octopus-worker-legacy.js: src/subtitles-octopus-worker.bc src/polyfill.js src/pre-worker.js src/SubOctpInterface.js src/post-worker.js build/lib/brotli/js/decode.js build/lib/brotli/js/polyfill.js
 	mkdir -p dist/js
 	emcc src/subtitles-octopus-worker.bc $(OCTP_DEPS) \
 		--pre-js src/polyfill.js \
+		--pre-js build/lib/brotli/js/polyfill.js \
 		--pre-js src/pre-worker.js \
 		--pre-js build/lib/brotli/js/decode.js \
 		--post-js src/SubOctpInterface.js \
--- a/Makefile
+++ b/Makefile
@@ -325,9 +325,10 @@ dist/js/subtitles-octopus-worker.js: src/subtitles-octopus-worker.bc src/pre-wor
 		-s WASM=1 \
 		$(EMCC_COMMON_ARGS)
 
-dist/js/subtitles-octopus-worker-legacy.js: src/subtitles-octopus-worker.bc src/pre-worker.js src/SubOctpInterface.js src/post-worker.js build/lib/brotli/js/decode.js
+dist/js/subtitles-octopus-worker-legacy.js: src/subtitles-octopus-worker.bc src/polyfill.js src/pre-worker.js src/SubOctpInterface.js src/post-worker.js build/lib/brotli/js/decode.js
 	mkdir -p dist/js
 	emcc src/subtitles-octopus-worker.bc $(OCTP_DEPS) \
+		--pre-js src/polyfill.js \
 		--pre-js src/pre-worker.js \
 		--pre-js build/lib/brotli/js/decode.js \
 		--post-js src/SubOctpInterface.js \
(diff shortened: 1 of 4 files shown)


TheOneric:
> Some thoughts
> Maybe extract polyfills in polyfills.js and add them only to the legacy worker?

Seems sensible, though if JSO switches to emscripten's [combined `WASM=2` mode](https://github.com/emscripten-core/emscripten/blob/fac7c5db1e755b67dbb89d7bf01b9d98e06532f8/src/settings.js#L1226-L1237) in the future it will need to be added to both again.

> Some of them (marked From brotli) are from https://github.com/google/brotli/blob/master/js/polyfill.js
> [...]  it is probably better to rely on Emscripten polyfills and explicitly added JSO polyfills.

I agree that having the polyfills provided by Emscripten would be ideal, but failing that utilisng brotli's `polyfill.js` seems better imho, as it should in theory automatically add polyfills only required for brotli whenever we update brotli — otherwise we would need to update this manually. If you're worried about brotli's polyfills not being general enough, loading our own polyfills before brotli's `polyfill.js` will result in our versions being used, right?

> For example, there is no Int8Array.slice in the version currently used by JSO, but there is in the latest recent.

Grepping for `\.slice` yielded no match in the currently used version of brotli's `decode.js`, so I guess this version of brotli doesn't yet need it. If it's already required we could also bump the brotli dependency beyond the current release.

dmitrylyzo:
> If you're worried about brotli's polyfills not being general enough, loading our own polyfills before brotli's polyfill.js will result in our versions being used, right?

I'm mostly concerned about polyfill duplication. No harm, but it will bloat the output file.

> Grepping for \.slice yielded no match in the currently used version of brotli's decode.js, so I guess this version of brotli doesn't yet need it. If it's already required we could also bump the brotli dependency beyond the current release.

It seems that `slice` is used by Emscripten: the `buffer.slice(offset,offset+length)` call in `MEMFS`.
Above, I meant that what if brotli suddenly stops using `slice` and removes the `slice` polyfill, it needs to be added back to cover Emscripten's needs.

Btw, in webOS 1.2 [`from` polyfill](https://github.com/google/brotli/blob/62662f87cdd96deda90ac817de94e3c4af75226a/js/polyfill.js#L1-L15) adds `from` to each `Object`, because `Int32Array.__proto__ === {}.__proto__: true`. It works, but seems wrong to me.

TheOneric:
> Above, I meant that what if brotli suddenly stops using slice and removes the slice polyfill, it needs to be added back to cover Emscripten's needs.

If it's used by emscripten not just brotli, than it would be defined in our own polyfills, which should take precedence over brotli's polyfills, and since it wouldn't be removed (until emscripten starts to provide the polyfills).
The intended benefit of including brotli polyfills lies in it automatically covering all needed polyfills for brotli, without anyone needing to test specifically old engines eacht ime we bump it. But ofc, there's also the possibilty that changes to JSO code or bumping emscripten might lead to new polyfills being required (and we don't test for that); perhaps this means not including brotli's own polyfills doesn't make much of a difference. Is the “bloat” and adding “`from` to each `Object`” problematic?

On another note, just copying polyfills from `lib/brotli/js/polyfills.js` to `src/…` can in gneral have licence/copyright-implications. By chance JSO and brotli use the same Expat-style "MIT"-licence, but we'd also need to copy the original copyright notice — but brotli's `polyfill.js` doesn't have one, so I'm not sure if we're supposed to fallback to the default attribtuion from brotli's `LICENCE` or just not do anything. Ofc, trivial code which the current ones *maybe* count as isn't copyrightable in the first place.

Makefile
```diff
@@ -325,9 +326,10 @@ dist/js/subtitles-octopus-worker.js: src/subtitles-octopus-worker.bc src/pre-wor
 		-s WASM=1 \
 		$(EMCC_COMMON_ARGS)
 
-dist/js/subtitles-octopus-worker-legacy.js: src/subtitles-octopus-worker.bc src/pre-worker.js src/SubOctpInterface.js src/post-worker.js build/lib/brotli/js/decode.js
+dist/js/subtitles-octopus-worker-legacy.js: src/subtitles-octopus-worker.bc src/pre-worker.js src/SubOctpInterface.js src/post-worker.js build/lib/brotli/js/decode.js build/lib/brotli/js/polyfill.js
 	mkdir -p dist/js
 	emcc src/subtitles-octopus-worker.bc $(OCTP_DEPS) \
+		--pre-js build/lib/brotli/js/polyfill.js \
 		--pre-js src/pre-worker.js \
 		--pre-js build/lib/brotli/js/decode.js \
```
TheOneric, line 334: I guess with our polyfills being merged into `src/pre-worker.js` we need to add `build/lib/brotli/js/polyfill.js` first so we can use the polyfills from brotli — but this means brotli's polyfill versions take precendence over ours. Can you prepend a commit which moves our polyfills into a separate file as you propsed (being only used by the legacy worker, if that's a safe assumption)? Then we can add brotli polyfills after ours.

dmitrylyzo: Only these? https://github.com/libass/JavascriptSubtitlesOctopus/blob/f045d1cf7bcda011d3306b4dd350b22c80a0c4f8/src/pre-worker.js#L1-L50

Or also this? https://github.com/libass/JavascriptSubtitlesOctopus/blob/f045d1cf7bcda011d3306b4dd350b22c80a0c4f8/src/pre-worker.js#L188-L203

TheOneric: Both.

TheOneric (approved):

PR #120, opened 2022-01-29
[Jellyfin] Rework buffer into C++ class

dmitrylyzo:
Cherry-picked from: https://github.com/jellyfin/JavascriptSubtitlesOctopus/commit/6d3a5c731f52ea84282e40bed3014ef95220146a
Cherry-picked from (part of): https://github.com/jellyfin/JavascriptSubtitlesOctopus/commit/0cc2bf8556f4936a483c8a027e0b2ff2774ed9d1

This is the only thing I can extract with such strict requirements in your repo. :neutral_face:
_* and my own principles of porting the results of teamwork from the fork._

### DIFF
--- a/src/SubtitleOctopus.cpp
+++ b/src/SubtitleOctopus.cpp
@@ -21,11 +21,11 @@ int log_level = 3;
 
 typedef struct {
     void *buffer;
-    int size;
-    int lessen_counter;
+    size_t size;
+    size_t lessen_counter;
 } buffer_t;
 
-void* buffer_resize(buffer_t *buf, int new_size, int keep_content) {
+void* buffer_resize(buffer_t *buf, size_t new_size, int keep_content) {
     if (buf->size >= new_size) {
         if (buf->size >= 1.3 * new_size) {
             // big reduction request
@@ -56,7 +56,7 @@ void* buffer_resize(buffer_t *buf, int new_size, int keep_content) {
 
 void buffer_init(buffer_t *buf) {
     buf->buffer = NULL;
-    buf->size = -1;
+    buf->size = 0;
     buf->lessen_counter = 0;
 }
 
--- a/src/SubtitleOctopus.cpp
+++ b/src/SubtitleOctopus.cpp
@@ -268,12 +268,13 @@ public:
         }
 
         // make float buffer for blending
-        float* buf = (float*)buffer_resize(&m_blend, sizeof(float) * width * height * 4, 0);
+        const size_t buffer_size = width * height * 4 * sizeof(float);
+        float* buf = (float*)buffer_resize(&m_blend, buffer_size, 0);
         if (buf == NULL) {
             fprintf(stderr, "jso: cannot allocate buffer for blending\n");
             return &m_blendResult;
         }
-        memset(buf, 0, sizeof(float) * width * height * 4);
+        memset(buf, 0, buffer_size);
 
         // blend things in
         for (cur = img; cur != NULL; cur = cur->next) {
--- a/src/SubtitleOctopus.cpp
+++ b/src/SubtitleOctopus.cpp
@@ -355,4 +355,6 @@ private:
 
 int main(int argc, char *argv[]) { return 0; }
 
+#ifdef __EMSCRIPTEN__
 #include "./SubOctpInterface.cpp"
+#endif
--- a/src/SubtitleOctopus.cpp
+++ b/src/SubtitleOctopus.cpp
@@ -19,50 +19,59 @@
 
 int log_level = 3;
 
-typedef struct {
+class ReusableBuffer {
+private:
     void *buffer;
     size_t size;
     size_t lessen_counter;
-} buffer_t;
 
-void* buffer_resize(buffer_t *buf, size_t new_size, int keep_content) {
-    if (buf->size >= new_size) {
-        if (buf->size >= 1.3 * new_size) {
-            // big reduction request
-            buf->lessen_counter++;
+public:
+    ReusableBuffer(): buffer(NULL), size(0), lessen_counter(0) {}
+
+    ~ReusableBuffer() {
+        free(buffer);
+    }
+
+    void clear() {
+        free(buffer);
+        buffer = NULL;
+        size = 0;
+        lessen_counter = 0;
+    }
+
+    void *ensure_size(size_t new_size, bool keep_content) {
+        if (size >= new_size) {
+            if (size >= 1.3 * new_size) {
+                // big reduction request
+                lessen_counter++;
+            } else {
+                lessen_counter = 0;
+            }
+            if (lessen_counter < 10) {
+                // not reducing the buffer yet
+                return buffer;
+            }
+        }
+
+        void *newbuf;
+        if (keep_content) {
+            newbuf = realloc(buffer, new_size);
         } else {
-            buf->lessen_counter = 0;
+            newbuf = malloc(new_size);
         }
-        if (buf->lessen_counter < 10) {
-            // not reducing the buffer yet
-            return buf->buffer;
-        }
+        if (!newbuf) return NULL;
+
+        if (!keep_content) free(buffer);
+        buffer = newbuf;
+        size = new_size;
+        lessen_counter = 0;
+        return buffer;
     }
 
-    void *newbuf;
-    if (keep_content) {
-        newbuf = realloc(buf->buffer, new_size);
-    } else {
-        newbuf = malloc(new_size);
+    size_t capacity() const {
+        return size;
     }
-    if (!newbuf) return NULL;
-
-    if (!keep_content) free(buf->buffer);
-    buf->buffer = newbuf;
-    buf->size = new_size;
-    buf->lessen_counter = 0;
-    return buf->buffer;
-}
-
-void buffer_init(buffer_t *buf) {
-    buf->buffer = NULL;
-    buf->size = 0;
-    buf->lessen_counter = 0;
-}
-
-void buffer_free(buffer_t *buf) {
-    free(buf->buffer);
-}
+};
 
 void msg_callback(int level, const char *fmt, va_list va, void *data) {
     if (level > log_level) // 6 for verbose
@@ -131,7 +140,7 @@ public:
         resizeCanvas(frame_w, frame_h);
 
         reloadFonts();
-        buffer_init(&m_blend);
+        m_blend.clear();
     }
 
     /* TRACK */
@@ -177,7 +186,7 @@ public:
         ass_free_track(track);
         ass_renderer_done(ass_renderer);
         ass_library_done(ass_library);
-        buffer_free(&m_blend);
+        m_blend.clear();
     }
     void reloadLibrary() {
         quitLibrary();
@@ -269,7 +278,7 @@ public:
 
         // make float buffer for blending
         const size_t buffer_size = width * height * 4 * sizeof(float);
-        float* buf = (float*)buffer_resize(&m_blend, buffer_size, 0);
+        float* buf = (float*)m_blend.ensure_size(buffer_size, false);
         if (buf == NULL) {
             fprintf(stderr, "jso: cannot allocate buffer for blending\n");
             return &m_blendResult;
@@ -349,7 +358,7 @@ public:
     }
 
 private:
-    buffer_t m_blend;
+    ReusableBuffer m_blend;
     RenderBlendResult m_blendResult;
 };
 


src/SubtitleOctopus.cpp
```diff
@@ -337,10 +346,12 @@ class SubtitleOctopus {
...
-    buffer_t m_blend;
+    ReusableBuffer m_blend;
     RenderBlendResult m_blendResult;
 };
 
 int main(int argc, char *argv[]) { return 0; }
 
+#ifdef __EMSCRIPTEN__
 #include "./SubOctpInterface.cpp"
+#endif
```
TheOneric, line 357: First of all, unrelated to the buffer refactor.
Second: Are there some tests or some other use for compiling this without emscripten? If not, I see no use in adding the guard in the first place.

TheOneric (commented):
> [dimtrylyzo:] This is the only thing I can extract with such strict requirements in your repo.

There are also fixes for seemingly pre-existing issues like https://github.com/libass/JavascriptSubtitlesOctopus/pull/111/commits/026b74422b716daa09edb01dc2d9218f66346e2f  *(in the linked example, at first glance it appears like it only needs a polished commit message)*.

src/SubtitleOctopus.cpp
dmitrylyzo: It allows you to compile the file alone, iirc.

src/SubtitleOctopus.cpp
```diff
@@ -19,50 +19,59 @@
...
     }
-    if (!newbuf) return NULL;
 
-    if (!keep_content) free(buf->buffer);
-    buf->buffer = newbuf;
-    buf->size = new_size;
-    buf->lessen_counter = 0;
-    return buf->buffer;
-}
+    void *take(int new_size, bool keep_content) {
```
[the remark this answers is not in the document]
dmitrylyzo: I tend to `get` for some reason :man_shrugging: 
This is more like `reserve`, because the buffer size will not be smaller than `new_size`.

src/SubtitleOctopus.cpp
TheOneric: I know, but is there even any *useful* standalone program or tests you can build with this file? For some reason SubtitleOctopus.cpp has a main function, but all it does is to immediately exit, which is not remotely useful.  
If there is no use for compiling it without emscripten, then the guard is not necessary and even misleading as it suggest there is some use in compiling it without emscripten.
If, on the other hand, there are some hidden tests which use this standalone we should integrate them into our CI.

src/SubtitleOctopus.cpp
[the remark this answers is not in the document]
TheOneric: How about `request_resize`, `ensure_size` or `reserve_size`?
`get` is imho to confusing as many array and list types use `get`-function to access individual members.

src/SubtitleOctopus.cpp
dmitrylyzo: I just use `g++ SubtitleOctopus.cpp` to see if it compiles without having to run the full build process.

This has the same purpose: https://github.com/libass/JavascriptSubtitlesOctopus/blob/de8888fe7f3fe5e52e50ece1407b96b68a244b5f/src/SubtitleOctopus.cpp#L13-L18

src/SubtitleOctopus.cpp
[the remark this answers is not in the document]
dmitrylyzo: Of these 3 options, `ensure_size` is probably the best fit.

I also thought about `allocate`, but this may confuse about the need of `free`.

We can stick to the original name `resize` :man_shrugging: 

Furthermore, I am thinking of adding a `size` method to return the passed/calculated `new_size` so that it is not recalculated later (see #124).

src/SubtitleOctopus.cpp
TheOneric: `make -C src` would also do, but if it’s already partly guarded anyway .. oh well. At least mention in the commit-message that this is only to quickly check for compiler warnings/errors and there’s no use to the compiled program or otheriwse.

src/SubtitleOctopus.cpp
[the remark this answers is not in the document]
TheOneric: Let’s use `ensure_size` then.

dmitrylyzo:
The `fix the error message printing` commit can be extracted from the branch if you want to do a merge commit.
I can also split the `m_` prepending if you want a clearer diff.

TheOneric:
- `[…] (for testing only)` is still ambigous as it could refer to unit tests and imho the explanation is better placed into the description rather than the title.
 - I see no benefit in prepending `m_`
 - I would have placed `int` → `size_t` in the overflow fix branch and/or a more general Buffer overhaul (see below), but it's fine here too
 - all other messages (also errros) are all logged to stdout
 - rather than returning a internal pointer in `ensure_resize` and memsetting this, creating a new function to memset the buffer or memsetting the buffer when `keep_content=false`. (see below) For just refactoring the buffer struct into a class (the scope of this PR) it can be left as is for now
 
Atm `keep_content=false` doesn't make much sense; if it's set to `true` but the old data is not intened to be kept, the buffer will be filled with known leftover garbage. If it’s set to `false` a fresh buffer is allocated which is filled with unknown garbage. Renaming it to `zero_mem` or so and letting it do a memset or removing it an introducing a dedicated function to zero the buffer seems more sensible.  
The only reason I can think of for the framesize being re-ensured every frame (with a request counter), is that it’s a workaround for continuos resizes firing multiple requests in rapid succession. Ideally the resize should only happen on worker init and canvas resize (with any continous-resize-workarounds contained in the JS-side)

For just refactoring the struct into a class, hte (initial) scope of this PR, I think just the refactoring itself, with a renaming of `resize` to `ensure_size` for clarity is fine, but keeping also the “standalone” (with a better commit message) and optionally the “size_t” commits as a collateral is ok.
Eg (if without “size_t”) like this: [compile alone](https://github.com/TheOneric/JavascriptSubtitlesOctopus/commit/2618515ab1b49cff05a48e1586d50b32ecb972af) + [refactor to class](https://github.com/TheOneric/JavascriptSubtitlesOctopus/commit/9e00aa4dfa1897b48519fbb0d57596c0157ef4c9)

dmitrylyzo:
> * I see no benefit in prepending `m_`

As here: https://github.com/libass/JavascriptSubtitlesOctopus/blob/a5afc8319a3e222c02ae6a19298a1754a33501df/src/SubtitleOctopus.cpp#L347-L348
Or it could be an underscore at the end.
Originally I wanted to add the `size` method (will conflict with the `size` data member). But now it may be redundant if the size is saved by the caller of `ensure_size`.

> * I would have placed `int` → `size_t` in the overflow fix branch and/or a more general Buffer overhaul (see below), but it's fine here too

I am currently driven by [this idea](https://github.com/libass/JavascriptSubtitlesOctopus/pull/124#issuecomment-1030580047). In this case the overflow protection goes after the refactor.

> * all other messages (also errros) are all logged to stdout

Do you want them to be output in `stderr` as well?

> * rather than returning a internal pointer in `ensure_resize` and memsetting this, creating a new function to memset the buffer or memsetting the buffer when `keep_content=false`.

It may not be necessary to zero the memory (even with `keep_content=false`) because you will manually set the actual values. _If we are talking about a general-purpose buffer._

> Atm `keep_content=false` doesn't make much sense; if it's set to `true` but the old data is not intened to be kept, the buffer will be filled with known leftover garbage. If it’s set to `false` a fresh buffer is allocated which is filled with unknown garbage. Renaming it to `zero_mem` or so and letting it do a memset or removing it an introducing a dedicated function to zero the buffer seems more sensible.

The value of `keep_content` depends on the usage of the buffer:
- it is `false` if you are not interested in old data.
- it is `true` if you are working with the buffer as an array. _But this only makes sense for the appending._

"zero-mem" is a sub-option of the first. `ensure_size_zero`? I think external memsetting provides more flexibility.

> The only reason I can think of for the framesize being re-ensured every frame (with a request counter), is that it’s a workaround for continuos resizes firing multiple requests in rapid succession. Ideally the resize should only happen on worker init and canvas resize (with any continous-resize-workarounds contained in the JS-side)

Originally maybe, but now smart blend uses 9 such buffers: https://github.com/jellyfin/JavascriptSubtitlesOctopus/blob/f4625ac313b318bd5d2e0ae18679ff516370bae6/src/SubtitleOctopus.cpp#L98-L103
They may have a different size on each frame, but their "capacity" will only grow to a certain value.

dmitrylyzo:
Currently, if #124 is fine, it should go first.

TheOneric:
The “Allow to compile CPP-file alone” commit has been merged.

> Currently, if https://github.com/libass/JavascriptSubtitlesOctopus/pull/124 is fine, it should go first.

Using some custom magic constant for the canvas dimensions is not the right way to prevent typewidth overflows.

I have an [untested branch](https://github.com/TheOneric/JavascriptSubtitlesOctopus/commits/buffer_refactor) started several weeks ago, which would replace both this and the overflow PR, but hadn't the time to finish + test it yet. Can you clarify what requirements are actually necessary for concretly planned or in jellyfin already existing features?

dmitrylyzo:
> Using some custom magic constant for the canvas dimensions is not the right way to prevent typewidth overflows.

But you can trim the canvas to a reasonable size.
Checking the size on each frame may degrade performance.
And `How hard do we need to defend ourselves against libass?`

> I have an [untested branch](https://github.com/TheOneric/JavascriptSubtitlesOctopus/commits/buffer_refactor) started several weeks ago, which would replace both this and the overflow PR, but hadn't the time to finish + test it yet. Can you clarify what requirements are actually necessary for concretly planned or in jellyfin already existing features?

1. [Here](https://github.com/jellyfin/JavascriptSubtitlesOctopus/blob/f4625ac313b318bd5d2e0ae18679ff516370bae6/src/SubtitleOctopus.cpp#L633) we need to know the size before allocation.
_This is `smart-blend` part._
2. The requested size may be less than `capacity` - [zeroing by capacity](https://github.com/TheOneric/JavascriptSubtitlesOctopus/blob/5285167f78f961f261e404cdb78767a39aad278a/src/SubtitleOctopus.cpp#L293) is more than necessary.

TheOneric:
Type overflows have nothing to do with libass or any notion of a "reasonable" canvas size *(if you want to do the latter there’s an JSO option for this)*, but with the sizes of the underlying architecture and how those are mapped to C types. Any custom magic constant is incorrect in that it either is below the minimum size required by ISO and may needlessly limit, or it is above the minimum and may not actually prevent overflows.
The only correct way to guard against type overflows is using the values from `limits.h`.

> 1. [Here](https://github.com/jellyfin/JavascriptSubtitlesOctopus/blob/f4625ac313b318bd5d2e0ae18679ff516370bae6/src/SubtitleOctopus.cpp#L633) we need to know the size before allocation.

You obviously still know what size you need beforehand, you just might need to do an overflow check before — or since `size_t` is unsigned, after if possible — blindly multiplying values.

> 2. The requested size may be less than capacity - [zeroing by capacity](https://github.com/TheOneric/JavascriptSubtitlesOctopus/blob/5285167f78f961f261e404cdb78767a39aad278a/src/SubtitleOctopus.cpp#L293) is more than necessary.

We could instead add an option to query a zero-initialised buffer from `get_rawbuf`. Or just recalculate the actual size, if `get_rawbuf` succeeded the multiplication is safe and I doubt this multiplication is in any under real-world conditions measurable way performance-relevant.

dmitrylyzo:
> Type overflows have nothing to do with libass

Those `width` and `height` (which initiated the overflow problem) are from libass, no?
Does libass protect itself from overflow due to resolution?

I ask because if you know the range of data you're working with, you don't have to get too defensive.
You don't create a public class that can be used by anyone but you.

> or any notion of a "reasonable" canvas size (if you want to do the latter there’s an JSO option for this), but with the sizes of the underlying architecture and how those are mapped to C types. Any custom magic constant is incorrect in that it either is below the minimum size required by ISO and may needlessly limit, or it is above the minimum and may not actually prevent overflows.

If you don't limit them, you will constantly get memory allocation errors.
Yes, you can check as you did in your branch, but 4GB (for 32 bit) is too much for a framebuffer, imo.
I think it's better to fix them or throw an error from the beginning.

> You obviously still know what size you need beforehand

Then the size will be calculated twice: outside the buffer and inside the buffer.

> Or just recalculate the actual size, if `get_rawbuf` succeeded the multiplication is safe and I doubt this multiplication is in any under real-world conditions measurable way performance-relevant.

Little by little such immeasurable places can grow into measurable one.
And, again, you will have a size calculation in two places.

TheOneric:
> Those width and height (which initiated the overflow problem) are from libass, no?

No. And the overflow happens in pure C.

PR #121, opened 2022-01-31
Update defaults

TheOneric:
Lower default fps to 24 and use WASM-side blending by default.


TheOneric:
If someone wishes to take a closer look at this and needs more time, say(/write) so now. If there are no replies I'll merge this in ~2 days.

PR #123, opened 2022-01-31
Fix blend render returning empty result

dmitrylyzo:
libass may return empty images, and 'blendResult' may become invalid (with zero height), leading to a runtime error when displaying.
Noticed on 'Railgun S Karaoke Test' ~26.38s (after the first verse) with `dropAllAnimations` enabled.

Would checking `<= 0` be better, since `width` and `height` are `int`?
They probably can't be negative by design, but... :thinking: 

This seems to be reproduced with `wasm-blend` on [empty_image.ass.txt](https://github.com/libass/JavascriptSubtitlesOctopus/files/7998336/empty_image.ass.txt)

### DIFF
--- a/src/SubtitleOctopus.cpp
+++ b/src/SubtitleOctopus.cpp
@@ -247,6 +247,7 @@ public:
         int max_x = img->dst_x + img->w - 1, max_y = img->dst_y + img->h - 1;
         ASS_Image *cur;
         for (cur = img->next; cur != NULL; cur = cur->next) {
+            if (cur->w == 0 || cur->h == 0) continue; // skip empty images
             if (cur->dst_x < min_x) min_x = cur->dst_x;
             if (cur->dst_y < min_y) min_y = cur->dst_y;
             int right = cur->dst_x + cur->w - 1;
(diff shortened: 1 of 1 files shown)


TheOneric:
Thanks! Can you be more verbose in your description on *why and how* exactly a zero width image leads to runtime errors later?

> Would checking <= 0 be better, since width and height are int?
> They probably can't be negative by design, but... 🤔

Internally the positions (but not the extnents afaik) can be negative, but before the images are returned to the API-user libass clips them to the positive range. Negative values being returned would be a libass bug and this is also [relied upon by mpv](https://github.com/mpv-player/mpv/blob/eca78ee9259bd477e6a8073165e3ebe22bc4423e/sub/ass_mp.c#L363-L386).

dmitrylyzo:
`Uncaught DOMException: Failed to construct 'ImageData': The source height is zero or not a number.`

TheOneric:
> `Uncaught DOMException: Failed to construct 'ImageData': The source height is zero or not a number.`
> 
> [In the commit message:] [...] leading to a runtime error when displaying, being unable to create an image with zero or invalid size.

I think the commit message could still be a bit more explicit naming what API can't create zero-area images, e.g.:
```
libass may return empty images, but browsers do not allow
creation of zero-area ImageData objects. Skipping and
exiting early on empty images avoids this and as a bonus
also saves a few CPU-cycles.

Initially noticed on the 'Railgun S Karaoke Test' sample
at ~26.38s (after the first verse) with `dropAllAnimations`
from jellyfin enabled.
```
Otherwise this seems good to merge to me, thanks.

PR #124, opened 2022-01-31
Framebuffer overflow protection

dmitrylyzo:
Quick fix for https://github.com/libass/JavascriptSubtitlesOctopus/pull/120#discussion_r795268373
> Pre-existing issue (so new commit to fix), but this line has potentially a big issue: Afaict there's no prior checks for this and the multiplications for the `new_size` parameter can overflow the bounds of `int`, which is in itself already undefined behaviour, but if your lucky only truncates. Worse though, the buffer is later accessed by values recalculated from `width` and `height`, so if the multiplication here overflowed (and the overflow itself doesn't blow up), an out-of-bound write happens either blowing up or corrupting some other memory. Is `ReusableBuffer` ever used for something else than an float-RGBA canvas alter down the line in jellyfin-patches? If not it's probably best to pass `width`, `height` and `member_size` as separate `size_t`-type values to the resize function and do all the overflow checking inside the function.

~Personally, I don't like re-calculation of size at this line `memset(buf, 0, sizeof(float) * width * height * 4);`~
_Stored in a variable before calling `buffer_resize`._

~I am thinking about:~
```cpp
typedef struct {
    float *buffer;
    size_t size;        // save the new_size
    size_t capacity;    // as the size member before
    size_t lessen_counter;
} buffer_t;
```
_Storing the `size` is not suitable for `ReusableBuffer`. But it is fine for  `FrameBuffer`._

### DIFF
--- a/src/SubtitleOctopus.cpp
+++ b/src/SubtitleOctopus.cpp
@@ -21,11 +21,11 @@ int log_level = 3;
 
 typedef struct {
     void *buffer;
-    int size;
-    int lessen_counter;
+    size_t size;
+    size_t lessen_counter;
 } buffer_t;
 
-void* buffer_resize(buffer_t *buf, int new_size, int keep_content) {
+void* buffer_resize(buffer_t *buf, size_t new_size, int keep_content) {
     if (buf->size >= new_size) {
         if (buf->size >= 1.3 * new_size) {
             // big reduction request
@@ -56,7 +56,7 @@ void* buffer_resize(buffer_t *buf, int new_size, int keep_content) {
 
 void buffer_init(buffer_t *buf) {
     buf->buffer = NULL;
-    buf->size = -1;
+    buf->size = 0;
     buf->lessen_counter = 0;
 }
 
--- a/src/SubtitleOctopus.cpp
+++ b/src/SubtitleOctopus.cpp
@@ -268,12 +268,13 @@ public:
         }
 
         // make float buffer for blending
-        float* buf = (float*)buffer_resize(&m_blend, sizeof(float) * width * height * 4, 0);
+        const size_t buffer_size = width * height * 4 * sizeof(float);
+        float* buf = (float*)buffer_resize(&m_blend, buffer_size, 0);
         if (buf == NULL) {
             fprintf(stderr, "jso: cannot allocate buffer for blending\n");
             return &m_blendResult;
         }
-        memset(buf, 0, sizeof(float) * width * height * 4);
+        memset(buf, 0, buffer_size);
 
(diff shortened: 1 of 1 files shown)


dmitrylyzo:
Changing `buffer_resize` in this way makes `keep_content` unnecessary, since it becomes a framebuffer (image) instead of a general-purpose buffer.

But, as I said before, a raw dynamic buffer may be needed in the future, e.g. to replace [m_is_event_animated](https://github.com/jellyfin/JavascriptSubtitlesOctopus/blob/f4625ac313b318bd5d2e0ae18679ff516370bae6/src/SubtitleOctopus.cpp#L567-L572).
_A template dynamic array would be better if it had no overhead._

So it might be better to refactor this in class first. Then wrap it with another `FrameBuffer` class, like:
```cpp
class FrameBuffer {
private:
    ReusableBuffer m_buffer;

public:
    float* ensure_size(size_t width, size_t height) {
        if (width == 0 || width > FRAMEBUFFER_MAX_WIDTH || height == 0 || height > FRAMEBUFFER_MAX_HEIGHT) {
            return NULL;
        }

        const size_t new_size = width * height * 4 * sizeof(float);

        return (float*)m_buffer.ensure_size(new_size, false);
    }

    size_t size() const {
         return m_buffer.size();
    }
};

**Comment by dmitrylyzo** [comment]
Because of [this](https://github.com/jellyfin/JavascriptSubtitlesOctopus/blob/f4625ac313b318bd5d2e0ae18679ff516370bae6/src/SubtitleOctopus.cpp#L633) it will be difficult to use the mentioned `FrameBuffer` class: need to know the size before use and `unsigned int` instead of `float`.

Would it be sufficient to check the size in the `resizeCanvas`?
How hard do we need to defend ourselves against `libass`? _Assuming `libass` won't return out-of-range images._
The downside: when I tried 200x200 as a max, the subtitles wasn't scaled because the main thread doesn't know about the trimming.

File: src/SubtitleOctopus.cpp [hunk 1]
```diff
@@ -159,6 +164,13 @@ class SubtitleOctopus {
 
     /* CANVAS */
     void resizeCanvas(int frame_w, int frame_h) {
+        if (frame_w > FRAMEBUFFER_MAX_WIDTH || frame_h > FRAMEBUFFER_MAX_HEIGHT) {
```
dmitrylyzo, line 167: Hmm, how about `<= 0`? Exit with some code or set some default values?
Also, portrait orientation :thinking:

TheOneric:
Using custom magic constant to guard against type overflows is incorrect. Superseded by #133.

PR #125, opened 2022-02-07
Fix error output

dmitrylyzo:
Maybe also remove `libass: ` from those not from `libass`?
https://github.com/libass/JavascriptSubtitlesOctopus/blob/8b1014b921f9c73187c59fcdac320d344ca59e0a/src/SubtitleOctopus.cpp#L237
https://github.com/libass/JavascriptSubtitlesOctopus/blob/8b1014b921f9c73187c59fcdac320d344ca59e0a/src/SubtitleOctopus.cpp#L277

### DIFF
--- a/src/SubtitleOctopus.cpp
+++ b/src/SubtitleOctopus.cpp
@@ -112,7 +112,7 @@ public:
     void initLibrary(int frame_w, int frame_h) {
         ass_library = ass_library_init();
         if (!ass_library) {
-            printf("ass_library_init failed!\n");
+            fprintf(stderr, "ass_library_init failed!\n");
             exit(2);
         }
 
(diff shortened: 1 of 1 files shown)


TheOneric:
`ass_utils.h` is an internal header and not installed to the system; including it outside of libass is not expected to work.
I've added `jso` prefixes and touched up the commit messages a bit here: https://github.com/TheOneric/JavascriptSubtitlesOctopus/commits/error-output. Does this seem alright @dmitrylyzo ?

dmitrylyzo:
> Does this seem alright @dmitrylyzo ?

Verbose but fine :wink:

TheOneric:
Merged as e6eee9ea7485969db8af2260f05eba4fcba676d2 + dfada30df1ab19e68671930244542dde5be55763 and implemented the suggested prefix change as 882fa5aa9c078859dc911f8e281f799593950e00.

PR #127, opened 2022-02-14
Add options for bounding and scaling the rendering canvas

TheOneric:
Cleaned and simplified (keeping the same results for positive saling values) version of https://github.com/libass/JavascriptSubtitlesOctopus/pull/119 backporting a jellyfin feature.
( @dmitrylyzo )

In particular I'd like to get some feedback on the name for the option added in the second commit: While I find `hardHeightLimit` to be an ok name, I was also considering `canvasHeightLimit`, `maxCanvasHeight` and `maxRenderingHeight` but I'm not sure if eg the `canvas`-names might cause confusion with the HTML-`<canvas>` element.

Another previously discussed topic was if I'd be better to bundle the `prescale` options into one object internally or also externally. I have no particular preference with regards to that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,11 @@ When creating an instance of SubtitleOctopus, you can set the following options:
                        (Default: `0` - no limit)
 - `libassGlyphLimit`: libass glyph cache memory limit in MiB (approximate)
                       (Default: `0` - no limit)
+- `prescaleFactor`: Scale down (`< 1.0`) the subtitles canvas to improve
+                    performance at the expense of quality, or scale it up (`> 1.0`).
+                    (Default: `1.0` - no scaling; must be a number > 0)
+- `prescaleHeightLimit`: The height beyond which the subtitles canvas won't be prescaled.
+                         (Default: `1080`)
 
 ### Rendering Modes
 #### JS Blending
--- a/README.md
+++ b/README.md
@@ -137,6 +137,9 @@ When creating an instance of SubtitleOctopus, you can set the following options:
                     (Default: `1.0` - no scaling; must be a number > 0)
 - `prescaleHeightLimit`: The height beyond which the subtitles canvas won't be prescaled.
                          (Default: `1080`)
+- `maxRenderHeight`: The maximum rendering height of the subtitles canvas.
+                     Beyond this subtitles will be upscaled by the browser.
+                     (Default: `0` - no limit)
 
 ### Rendering Modes
 #### JS Blending
--- a/src/subtitles-octopus.js
+++ b/src/subtitles-octopus.js
@@ -17,6 +17,8 @@ var SubtitlesOctopus = function (options) {
     self.libassMemoryLimit = options.libassMemoryLimit || 0;
     self.libassGlyphLimit = options.libassGlyphLimit || 0;
     self.targetFps = options.targetFps || 24;
+    self.prescaleFactor = options.prescaleFactor || 1.0;
+    self.prescaleHeightLimit = options.prescaleHeightLimit || 1080;
     self.isOurCanvas = false; // (internal) we created canvas and manage it
     self.video = options.video; // HTML video element (optional if canvas specified)
     self.canvasParent = null; // (internal) HTML canvas parent element
@@ -404,14 +406,36 @@ var SubtitlesOctopus = function (options) {
         }
     };
 
+    function _computeCanvasSize(width, height) {
+        var scalefactor = self.prescaleFactor <= 0 ? 1.0 : self.prescaleFactor;
+
+        if (height <= 0 || width <= 0) {
+            width = 0;
+            height = 0;
+        } else {
+            var sgn = scalefactor < 1 ? -1 : 1;
+            var newH = height;
+            if (sgn * newH * scalefactor <= sgn * self.prescaleHeightLimit)
+                newH *= scalefactor;
+            else if (sgn * newH < sgn * self.prescaleHeightLimit)
+                newH = self.prescaleHeightLimit;
+
+            width *= newH / height;
+            height = newH;
+        }
+
+        return {'width': width, 'height': height};
+    }
+
     self.resize = function (width, height, top, left) {
         var videoSize = null;
         top = top || 0;
         left = left || 0;
         if ((!width || !height) && self.video) {
             videoSize = self.getVideoPosition();
-            width = videoSize.width * self.pixelRatio;
-            height = videoSize.height * self.pixelRatio;
+            var newSize = _computeCanvasSize(videoSize.width * self.pixelRatio, videoSize.height * self.pixelRatio);
+            width = newSize.width;
+            height = newSize.height;
             var offset = self.canvasParent.getBoundingClientRect().top - self.video.getBoundingClientRect().top;
             top = videoSize.y - offset;
             left = videoSize.x;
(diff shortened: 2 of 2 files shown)


TFSThiagoBR98:
> In particular I'd like to get some feedback on the name for the option added in the second commit: While I find `hardHeightLimit` to be an ok name, I was also considering `canvasHeightLimit`, `maxCanvasHeight` and `maxRenderingHeight` but I'm not sure if eg the `canvas`-names might cause confusion with the HTML-`<canvas>` element.

To me `maxCanvasHeight` or `maxRenderingHeight` looks good, I don't see any problems using canvas names as it uses canvas element for rendering, so I'll leave it up to you to decide the final name.

src/subtitles-octopus.js
```diff
@@ -404,14 +407,38 @@ var SubtitlesOctopus = function (options) {
         }
     };
 
+    function _computeCanvasSize(width, height) {
+        var scalefactor = self.prescaleFactor <= 0 ? 1.0 : self.prescaleFactor;
+
+        if (height == 0) {
```
TFSThiagoBR98, line 413: I think it's better to check the width here as well

`if (height == 0 || width == 0) {`

TheOneric: It's not strictly required for `_computeCanvas` to work, but sure, normalising all zero and negative cases to `(0, 0))`sounds good.
Now using `if (height <= 0 || width <= 0) {`.

TheOneric:
> I don't see any problems using canvas names as it uses canvas element for rendering

I was thinking of possible confusion with the CSS `height` property of the HTML `<canvas>` being rendered on. Not sure how likely/problematic this is though.
For now I renmaed the property to `maxRenderHeight`.

TheOneric:
@dmitrylyzo: I don't think I changed any functionality jellyfin relies on during cleanup, but just to make sure could you take a look?

src/subtitles-octopus.js
```diff
@@ -404,14 +407,39 @@ var SubtitlesOctopus = function (options) {
...
+        var scalefactor = self.prescaleFactor <= 0 ? 1.0 : self.prescaleFactor;
+
+        if (height <= 0 || width <= 0) {
+            width = 0;
+            height = 0;
+        } else {
+            var sgn = scalefactor < 1 ? -1 : 1;
+            var newH = height;
+            if (sgn * newH * scalefactor <= sgn * self.prescaleHeightLimit)
+                newH *= self.prescaleFactor;
```
dmitrylyzo, line 420:
```suggestion
                newH *= scalefactor;
```

src/subtitles-octopus.js
```diff
@@ -404,14 +407,39 @@ var SubtitlesOctopus = function (options) {
...
+    function _computeCanvasSize(width, height) {
+        var scalefactor = self.prescaleFactor <= 0 ? 1.0 : self.prescaleFactor;
+
+        if (height <= 0 || width <= 0) {
+            width = 0;
+            height = 0;
+        } else {
+            var sgn = scalefactor < 1 ? -1 : 1;
+            var newH = height;
+            if (sgn * newH * scalefactor <= sgn * self.prescaleHeightLimit)
```
dmitrylyzo, line 419: Given your desire to add protective checks, you could check `self.prescaleHeightLimit <= 0`.

dmitrylyzo (commented):
Probably the only thing left is a true noop (skipping multiplications) when `scaleFactor` is `1.0`. But since this function is rarely used, noop doesn't make much sense.

Should we round the result (to pixels)?

README.md
```diff
@@ -132,6 +132,14 @@ When creating an instance of SubtitleOctopus, you can set the following options:
                        (Default: `0` - no limit)
 - `libassGlyphLimit`: libass glyph cache memory limit in MiB (approximate)
                       (Default: `0` - no limit)
+- `prescaleFactor`: Scale down (`< 1.0`) the subtitles canvas to improve
+                    performance at the expense of quality, or scale it up (`> 1.0`).
```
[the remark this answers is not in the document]
TheOneric: Well.. depending on the content and the downscaling algorithm used *(which we don't control but is chosen by the browser-engine)* rendering at a higher resolution+downscale might not necessarily improve quality or at worst even decrease it by e.g. introducing downscaling artefacts like aliasing.

dmitrylyzo:
> `1.0` **already is** a “true” noop. Afaict for all IEEE-754 compliant floating point implementations, if `a` represents a real number, then `a * 1.0 === a` and if `a` is additionally non-zero `a / a === 1.0`. Meaning if `scalefactor === 1.0` the passed width and height values will be returned unchanged — a noop.

I meant that all this math will be applied even with `scalefactor === 1.0`, but could be omitted ("true noop").
But...
> since this function is rarely used, noop doesn't make much sense.

TheOneric:
Yeah, `_computeCanvas` is already lightweight and only runs on resizes and init, so there's no real performance benefit to adding another branch to skip the remaining code.

ThaUnknown:
@dmitrylyzo @TheOneric are there any more JS PR's to merge which aren't already made? If not I could start work on v2 and have it finished by the end of the week.

dmitrylyzo (approved):

TheOneric:
> are there any more JS PR's to merge which aren't already made?

yes.

ThaUnknown:
> > are there any more JS PR's to merge which aren't already made?
> 
> yes.

sorry, i meant from the original pr Dimitry made, any chance you could abridge me what those are?

ThaUnknown:
> > from the original pr Dimitry made [i.e. pr#111], any chance you could abridge me what those are?
> 
> Excluding this PR there is:
> 
> Merged:
> 
> * fix handling of empty ASS_Images
> 
> Created but not final or not reviewed::
> 
> * conditionally redirect C++ output for nicer display in the browser console
> * Refactor buffer struct into C++-class
>   
>   * additionally: fixing newly found overflow bugs
> 
> Uncreated:
> 
> * drop animations feature
> * `renderAhead`
>   
>   * additionally: improvements to `renderAhead` to never be worse than wasm-blend (the state in PR#111 is worse when falling behind schedule)

Here's what I think:
- drop animations feature: this is really good for a lot of low-end devices and has good applications, so we should do it
- renderAhead: imho this idea is fundamentally flawed, it should render ahead/buffer measured in time, rather than memory, I plan on creating something like this, additionally some of its components like 'renderOnDemand' arent exactly what they describe, and I plan on creating a renderOnDemand mode

Additionally, ideas which aren't posted here:
- font improvements: JS & C++: I didn't understand their point in all honesty, but I didn't have to work with massive font files, so there might be valid reason for them
- onDemandRender: JS: render subtitles as the video player is presenting video frames using `requestVideoFrameCallback`, chromium only [currently working on this]
- offscreenRender: JS: handling rendering 100% in the worker, chromium only, firefox support soon:tm:
- C++ decode: the entire code for decoding bitmaps NEEDS to be in C++ rather than JS, blendrender already creates the bitmaps in C++, but it additionally blends them, I'd like an options for C++ to create multiple bitmaps without blending them, rather than creating 1 blended bitmap TLDR moving [this](https://github.com/libass/JavascriptSubtitlesOctopus/blob/master/src/post-worker.js#L288-L325) code to C++

requestVideoFrameCallback can potentially allow us to do frame-perfect painting of subtitles if renderahead is done with time, rather than memory.

TheOneric:
@ThaUnknown: this PR is about prescaling and bounding the canvas and not the right place for extended discussion of general ideas for the future.

Issue #97, opened 2021-05-15
Memory leak: Event listeners not removing


cawa-93:
After init, the JSO creates many event listeners, and none of them is removed even if i call `instance.dispose();`

This is very noticeable especially if you work with SPA: even when you go to another "page", and the target video element was removed from DOM, event listeners continue to work. 

I think this is what leads to a constant memory leak: None of the created instances of the class can be destroyed by the garbage collector. 
![зображення](https://user-images.githubusercontent.com/1662812/118362940-7f889c00-b581-11eb-9018-a6dbdb768ab7.png)
![зображення](https://user-images.githubusercontent.com/1662812/118362987-c080b080-b581-11eb-8285-efcf79e1bba5.png)

Issue #72, opened 2020-06-09
Extreme lag on heavy "move" elements in dialogue


nickfujita:
It seems that when there are `move` animations in the sub file dialog, it causes the renderer to lag heavily, and in some cases crash the worker. I was previously using version 2.2.0, but saw that there were recently some promising improvements with alternate render techniques `lossyRender` and `blendRender`. After trying the latest code with different combinations of both the lossy and blend render modes, I can't seem to get the rendering to improve.

While using the `blendRender`, I've also tried setting `dropAllAnimations` to true, and `targetFps` to `5`. While I am no longer seeing the working crashing, the `move` animations are not dropped, and the animation duration lags past the timeline position where they are supposed to disappear. I've also tried different combinations of setting `libassMemoryLimit` and `libassGlyphLimit` both lower (in hopes that it will reduce the resources being used and simply skip over heavy loads), and higher (in hopes that the increased resources would improve render performance).

Below is a sample of the dialog combination which is attempting to be rendered. Was wondering if @JustAMan or anyone else possibly had any recommendations on configurations to improve rendering performance, or at least allow the renderer to keep in sync with the timeline and just skip frames when under heavy load.

```
[Script Info]
Title: 
Original Script:
Original Translation: 
Original Editing: 
Original Timing: 
Synch Point: 
Script Updated By: 
Update Details: 
ScriptType: v4.00+
Collisions: Normal
PlayResX: 640
PlayResY: 360
Timer: 0.0000
WrapStyle: 0

[V4+ Styles]
Format: Name,Fontname,Fontsize,PrimaryColour,SecondaryColour,OutlineColour,BackColour,Bold,Italic,Underline,Strikeout,ScaleX,ScaleY,Spacing,Angle,BorderStyle,Outline,Shadow,Alignment,MarginL,MarginR,MarginV,Encoding
Style: Default,Arial,20,&H00FFFFFF,&H000000FF,&H00000000,&H00000000,0,0,0,0,100,100,0,0,1,2,2,2,0010,0010,0010,1
Style: Main,Trebuchet MS,24,&H00FFFFFF,&H000000FF,&H00000000,&H00000000,0,0,0,0,100,100,0,0,1,2,1,2,0010,0010,0018,1

[Events]
Format: Layer,Start,End,Style,Name,MarginL,MarginR,MarginV,Effect,Text
[...]

Issue #46, opened 2019-05-09
Browser freezing under 'heavy' subs


starkana:
I recently got JavascriptSubtitlesOctopus up and running and testest a few videos, most which work great, but I noticed a couple videos with heavy intro subs (animation and effects) overloads my browser and freezes. From my understanding it should be able to handle it without lag. Perhaps I need to adjust some configurations? Below is a screenshot of the console output log.

![Capture](https://user-images.githubusercontent.com/7329623/57481638-29810f80-7268-11e9-858c-bec8c5694b6b.PNG)

Yay295:
Could you share the subs you're having trouble with?

starkana:
@Yay295 Thanks for replying. Here is the ASS file along with the fonts used: https://drive.google.com/drive/folders/10li-vh6EO-0YfvK3GZqTW0Tqm1jfNI31?usp=sharing

starkana:
@Yay295 did you get the chance to look into it?

Yay295:
A bit, but I don't have a conclusion. It is a big file, but it doesn't actually have anything too complex in it, and you can see from [this demo](https://dador.github.io/JavascriptSubtitlesOctopus/railgun.html) that it can handle similar files (mostly) fine. I think it's probably just the sheer number of subtitle "lines" in the file. For the OP, some of the individual letters are composed of 50 or more lines to make it look how it does.

no1d:
In my experience, at least in chrome, freeze happens when cache grows too large (or too fast?). Seems like memory reallocations in emscripten are becoming too expensive. But I don't really know much about it.
And after freeze video will actually play ok.
One more thing. Usually only 1080p are affected, while 720p are perfectly fine - another evidence of memory hog.

no1d:
Scratch my previous comment, through maybe its a different issue.
Today I played with some heavy subs, and I noticed that they have large number of bitmaps for each frame.
So I tried to rewrite rendering part using `createImageBitmap` and got pretty good results.
Keep in mind, I'm not very familiar with this stuff.
I'm not sure if async breaks something, and there is no exception handling.

**subtitles-octopus.js**
```
function renderFrames() {
    var data = self.renderFramesData;
    var beforeDrawTime = performance.now();
    self.ctx.clearRect(0, 0, self.canvas.width, self.canvas.height);
    for (var i = 0; i < data.bitmaps.length; i++) {
        var image = data.bitmaps[i];
[...]

PR #128, opened 2022-02-20
rewrite: "v2"

ThaUnknown:
This is a full rewrite of the JS of JSSO, meaning it includes breaking changes, hence the 'v2'.

I have used ES2020, but that can easily be compiled down by the user with the likes of `babel` or other, I however didn't force any reliance on new API's, nor have I removed any backwards compatibility, which means this [hopefully, pls check @dmitrylyzo ] should still run on any legacy browsers [after compiling using babel]. Maybe we could additionally provide a version of JSSO which is already babel'ed so the user doesn't have to do it? That's another topic.

I changed a lot of things, which means more can be changed. If you think something should be further changed, please say so.

The README still needs to be updated, but that always comes last.

The reason it's a single commit is because it was a full rewrite, rather than bit by bit patching, and since there's a "no broken commits" policy here, this is the only way I could imagine of doing it.

In general this is focused on performance, performance, performance [where it matters] and bugfixes.

Overview:

Structure:
- now using classes, so the structure is a lot less chaotic
- now extending EventTarget
  - this allows the user to run `so.addEventListener` and `so.removeEventListener` rather than using `so.oncustomeventname`
  https://github.com/ThaUnknown/JavascriptSubtitlesOctopus/blob/v2/src/subtitles-octopus.js#L477
  https://github.com/ThaUnknown/JavascriptSubtitlesOctopus/blob/v2/src/subtitles-octopus.js#L543
- using standardjs as a linter, because seeing 2 different code styles, sometimes even in the same commit was painful
- using const, let, destructing, anon functions, null coales, and other  
- things the user shouldnt use are prefixed with `_` so they show up at the end of stuff like console inspect etc.
- using nodejs standards for things like private attribute naming `_`, callback styles etc
https://github.com/ThaUnknown/JavascriptSubtitlesOctopus/blob/v2/src/subtitles-octopus.js#L444
https://github.com/ThaUnknown/JavascriptSubtitlesOctopus/blob/v2/src/subtitles-octopus.js#L511
https://github.com/ThaUnknown/JavascriptSubtitlesOctopus/blob/v2/src/subtitles-octopus.js#L501

Features:
- removed "render mode" replaced by "blending mode: `wasm`/`js`"
  - this allows the user to "mix and match" rendering "modifiers" of which there are now a few, like for example: `wasm`, `async`,`offscreen` and `onDemand`, or `js`, `sync`, `main`, `event`
  https://github.com/ThaUnknown/JavascriptSubtitlesOctopus/blob/v2/src/post-worker.js#L165
  https://github.com/ThaUnknown/JavascriptSubtitlesOctopus/blob/v2/src/post-worker.js#L178
  - the default is now offscreen, ondemand, async, wasm rendering, which turns off parts of it if the browser lacks support
- made `asyncRender` a modifier, which means you can now use it with blendrender
- created `offscreenRender`, a modifier which allows you to handle all the rendering on the worker using OffscreenCanvases [chrome only, firefox very soon]
https://github.com/ThaUnknown/JavascriptSubtitlesOctopus/blob/v2/src/subtitles-octopus.js#L58
https://github.com/ThaUnknown/JavascriptSubtitlesOctopus/blob/v2/src/subtitles-octopus.js#L86-L87
https://github.com/ThaUnknown/JavascriptSubtitlesOctopus/blob/v2/src/post-worker.js#L253
- created `onDemandRender`, a modifier which renders subtitles as the video player decodes video frames, rather than relying on events, this makes it insanely reliable to paint subtitles [currently chrome only, firefox shows interest but isnt working on this yet, for browsers which don't support it, it's polyfilled]
https://github.com/ThaUnknown/JavascriptSubtitlesOctopus/blob/v2/src/subtitles-octopus.js#L444-L448
https://github.com/ThaUnknown/JavascriptSubtitlesOctopus/blob/v2/src/post-worker.js#L238
https://github.com/ThaUnknown/JavascriptSubtitlesOctopus/blob/v2/src/subtitles-octopus.js#L577

Performance/bugs:
- async/offscreen/ondemand implementation/changes
- JS blend decode is now a LOT faster than before, thanks to some uint32array and byteshift magic
https://github.com/ThaUnknown/JavascriptSubtitlesOctopus/blob/v2/src/post-worker.js#L213-L226
  - I however don't know how this will behave on devices whos endianness is different from x86, can some1 test this? 
  https://github.com/ThaUnknown/JavascriptSubtitlesOctopus/blob/v2/src/post-worker.js#L222
  - the whole decode needs to be done inside WASM, rather than on JS, this code is now a lot faster, but it still can be a lot faster if it was done inside WASM
- the browser will no longer freeze when a CSS animation resizes the canvas
https://github.com/ThaUnknown/JavascriptSubtitlesOctopus/blob/v2/src/subtitles-octopus.js#L170
  - for instant dimension changes [fullscreen change, or flat resize] canvas will update instantly
https://github.com/ThaUnknown/JavascriptSubtitlesOctopus/blob/v2/src/subtitles-octopus.js#L177-L182
  - for any dynamic dimension changes [resizing window, css animation, etc] JSSO will wait for the animation to finish before asking for a canvas re-draw
https://github.com/ThaUnknown/JavascriptSubtitlesOctopus/blob/v2/src/subtitles-octopus.js#L169-L175
- the canvas will no longer randomly disappear [removed resize with timeout which was a very lazy hack]
- the canvas will no longer loose content when changing position without changing dimensions
https://github.com/ThaUnknown/JavascriptSubtitlesOctopus/blob/v2/src/subtitles-octopus.js#L165
- when resizing, the content of the canvas doesn't disappear, just gets rescaled until the next frame is painted
https://github.com/ThaUnknown/JavascriptSubtitlesOctopus/blob/v2/src/subtitles-octopus.js#L177-L179
https://github.com/ThaUnknown/JavascriptSubtitlesOctopus/blob/v2/src/post-worker.js#L109-L112
- a few idiot checks for people making dumb mistakes
https://github.com/ThaUnknown/JavascriptSubtitlesOctopus/blob/v2/src/subtitles-octopus.js#L246
- attempt to fix high GPU VRAM usage caused by asyncRender's ImageBitmaps, by closing them, but this doesn't reduce it by all that much
https://github.com/ThaUnknown/JavascriptSubtitlesOctopus/blob/v2/src/subtitles-octopus.js#L456
https://github.com/ThaUnknown/JavascriptSubtitlesOctopus/blob/v2/src/post-worker.js#L259

Test link: https://thaunknown.github.io/JavascriptSubtitlesOctopus/ts

### DIFF
--- a/src/post-worker.js
+++ b/src/post-worker.js
@@ -1,760 +1,637 @@
-Module['FS'] = FS;
+/* global Module, HEAPU8, FS, SDL */
+/* eslint-env browser, worker */
+/* eslint node/no-callback-literal: 0 */
+Module.FS = FS
 
-self.delay = 0; // approximate delay (time of render + postMessage + drawImage), for example 1/60 or 0
-self.lastCurrentTime = 0;
-self.rate = 1;
-self.rafId = null;
-self.nextIsRaf = false;
-self.lastCurrentTimeReceivedAt = Date.now();
-self.targetFps = 24;
-self.libassMemoryLimit = 0; // in MiB
+self.delay = 0 // approximate delay (time of render + postMessage + drawImage), for example 1/60 or 0
+self.lastCurrentTime = 0
+self.rate = 1
+self.rafId = null
+self.nextIsRaf = false
+self.lastCurrentTimeReceivedAt = Date.now()
+self.targetFps = 24
+self.libassMemoryLimit = 0 // in MiB
 
-self.width = 0;
-self.height = 0;
+self.width = 0
+self.height = 0
 
-self.fontMap_ = {};
-self.fontId = 0;
+self.fontMap_ = {}
+self.fontId = 0
 
 /**
  * Make the font accessible by libass by writing it to the virtual FS.
  * @param {!string} font the font name.
  */
-self.writeFontToFS = function(font) {
-    font = font.trim().toLowerCase();
+self.writeFontToFS = function (font) {
+  font = font.trim().toLowerCase()
 
-    if (font.startsWith("@")) {
-        font = font.substr(1);
-    }
+  if (font.startsWith('@')) {
+    font = font.substring(1)
+  }
 
-    if (self.fontMap_.hasOwnProperty(font)) return;
+  if (self.fontMap_[font]) return
 
-    self.fontMap_[font] = true;
+  self.fontMap_[font] = true
 
-    if (!self.availableFonts.hasOwnProperty(font)) return;
-    var content = readBinary(self.availableFonts[font]);
+  if (!self.availableFonts[font]) return
+  const content = readBinary(self.availableFonts[font])
 
-    Module["FS"].writeFile('/fonts/font' + (self.fontId++) + '-' + self.availableFonts[font].split('/').pop(), content, {
-        encoding: 'binary'
-    });
-};
+  Module.FS.writeFile('/fonts/font' + (self.fontId++) + '-' + self.availableFonts[font].split('/').pop(), content, {
+    encoding: 'binary'
+  })
+}
 
 /**
  * Write all font's mentioned in the .ass file to the virtual FS.
  * @param {!string} content the file content.
  */
-self.writeAvailableFontsToFS = function(content) {
-    if (!self.availableFonts) return;
+self.writeAvailableFontsToFS = function (content) {
+  if (!self.availableFonts) return
 
-    var sections = parseAss(content);
+  const sections = parseAss(content)
 
-    for (var i = 0; i < sections.length; i++) {
-        for (var j = 0; j < sections[i].body.length; j++) {
-            if (sections[i].body[j].key === 'Style') {
-                self.writeFontToFS(sections[i].body[j].value['Fontname']);
-            }
-        }
+  for (let i = 0; i < sections.length; i++) {
+    for (let j = 0; j < sections[i].body.length; j++) {
+      if (sections[i].body[j].key === 'Style') {
+        self.writeFontToFS(sections[i].body[j].value.Fontname)
+      }
     }
+  }
 
-    var regex = /\\fn([^\\}]*?)[\\}]/g;
-    var matches;
-    while (matches = regex.exec(self.subContent)) {
-        self.writeFontToFS(matches[1]);
-    }
-};
-
-self.getRenderMethod = function () {
-    switch (self.renderMode) {
-        case 'lossy':
-            return self.lossyRender;
-        case 'js-blend':
-            return self.render;
-        default:
-            console.error('Unrecognised renderMode, falling back to default!');
-            self.renderMode = 'wasm-blend';
-            // fallthrough
-        case 'wasm-blend':
-            return self.blendRender;
-    }
+  const regex = /\\fn([^\\}]*?)[\\}]/g
+  let matches
+  while (matches = regex.exec(self.subContent)) {
+    self.writeFontToFS(matches[1])
+  }
 }
-
 /**
  * Set the subtitle track.
  * @param {!string} content the content of the subtitle file.
  */
-self.setTrack = function (content) {
-    // Make sure that the fonts are loaded
-    self.writeAvailableFontsToFS(content);
+self.setTrack = function ({ content }) {
+  // Make sure that the fonts are loaded
+  self.writeAvailableFontsToFS(content)
 
-    // Write the subtitle file to the virtual FS.
-    Module["FS"].writeFile("/sub.ass", content);
+  // Write the subtitle file to the virtual FS.
+  Module.FS.writeFile('/sub.ass', content)
 
-    // Tell libass to render the new track
-    self.octObj.createTrack("/sub.ass");
-    self.ass_track = self.octObj.track;
-    self.getRenderMethod()();
-};
+  // Tell libass to render the new track
+  self.octObj.createTrack('/sub.ass')
+  self.ass_track = self.octObj.track
+  self.renderLoop()
+}
 
 /**
  * Remove subtitle track.
  */
 self.freeTrack = function () {
-    self.octObj.removeTrack();
-    self.getRenderMethod()();
-};
+  self.octObj.removeTrack()
+  self.renderLoop()
+}
 
 /**
  * Set the subtitle track.
  * @param {!string} url the URL of the subtitle file.
  */
-self.setTrackByUrl = function (url) {
-    var content = "";
-    if (isBrotliFile(url)) {
-        content = Module["BrotliDecode"](readBinary(url))
-    } else {
-        content = read_(url);
-    }
-    self.setTrack(content);
-};
+self.setTrackByUrl = function ({ url }) {
+  let content = ''
+  if (isBrotliFile(url)) {
+    content = Module.BrotliDecode(readBinary(url))
+  } else {
+    content = read_(url)
+  }
+  self.setTrack({ content })
+}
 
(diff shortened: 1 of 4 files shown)


TFSThiagoBR98:
> I have used ES2020, but that can easily be compiled down by the user with the likes of babel or other, I however didn't force any reliance on new API's, nor have I removed any backwards compatibility, which means this [hopefully, pls check @dmitrylyzo ] should still run on any legacy browsers [after compiling using babel]. Maybe we could additionally provide a version of JSSO which is already babel'ed so the user doesn't have to do it? That's another topic.

Isn't it better to include babel and eslint as devDependencies and compile it along with the worker, instead of include the compiled version.

ThaUnknown:
> > I have used ES2020, but that can easily be compiled down by the user with the likes of babel or other, I however didn't force any reliance on new API's, nor have I removed any backwards compatibility, which means this [hopefully, pls check @dmitrylyzo ] should still run on any legacy browsers [after compiling using babel]. Maybe we could additionally provide a version of JSSO which is already babel'ed so the user doesn't have to do it? That's another topic.
> 
> Isn't it better to include babel and eslint as devDependencies and compile it along with the worker, instead of include the compiled version.

Most likely, yes. But that's outside of the scope of this PR, I'd rather focus on the functionality/code

ThaUnknown:
I put up a "small" test case for this, ~300-500 bitmaps per frame, on firefox this is likely to crash, runs fine on chrome tho
https://miru.kirdow.com/ please ignore the awful video quality:size

starkana:
@ThaUnknown subs do not show in fullscreen mode

ThaUnknown:
> @ThaUnknown subs do not show in fullscreen mode

use f11, not video controls

ThaUnknown:
updated example video with some CSS animation showcases

src/post-worker.js
```diff
@@ -226,15 +226,18 @@ self.decodeResultBitmap = ({ bitmap, stride, w, h, color, dst_x, dst_y }) => {
 }
 
 self.render = (time, force) => {
+  self.bussy = true
```
Yay295, line 229: "busy" only has one "s"

ThaUnknown:
@TFSThiagoBR98 @dmitrylyzo @TheOneric does anything need to be changed for this PR for it to be reviewed? At most I could add URLs to code snippets of the fixes/features if thats required

TFSThiagoBR98:
> Most likely, yes. But that's outside of the scope of this PR, I'd rather focus on the functionality/code

**If possible**, can you add JSDoc comments to subtitle-octopus.js? I'll use that to generate the definition files for #95.

> At most I could add URLs to code snippets of the fixes/features if thats required.
That may help

ThaUnknown:
> **If possible**, can you add JSDoc comments to subtitle-octopus.js? I'll use that to generate the definition files for #95.

this is some of the least fun stuff I've ever done

ThaUnknown:
> That may help

added code snippets to wherever it was possible

src/post-worker.js
```diff
@@ -1,758 +1,638 @@
...
+  }
+  if (calledMain && messageResenderTimeout) {
+    clearTimeout(messageResenderTimeout)
+    messageResender()
+  }
+  const data = message.data
+  if (self[data.target]) {
+    self[data.target](data)
+  } else {
+    switch (data.target) {
```
TFSThiagoBR98, line 491: You probably forgot to implement the freeTrack, setTrack and setTrackByUrl in the onmessage event

TFSThiagoBR98: Also init

src/post-worker.js
```diff
@@ -1,758 +1,638 @@
...
+        nextRAF += 1000 / self.targetFps
+      }
+    }
+    const delay = Math.max(nextRAF - now, 0)
+    return setTimeout(func, delay)
+    // return setTimeout(func, 1);
+  }
+})()
+
+const screen = {
```
TFSThiagoBR98, line 364: This causes a conflict with screen object, can you change the name to screenCanvas

```
building:ERROR: Closure compiler run failed:

building:ERROR: /tmp/emscripten_temp_a8fms20c/subtitles-octopus-worker-legacy.js.pp.js.mem.js.jso.js.jso.js.jso.js.jso.js:1:4095: ERROR - [JSC_REDECLARED_VARIABLE_ERROR] Illegal redeclared variable: screen
```

ThaUnknown: https://github.com/libass/JavascriptSubtitlesOctopus/blob/master/src/post-worker.js#L431
looks like my linter tripped up and changed this to const, I assume emscripten re-defines this, which is weird cuz I expected this to be treated like a global

src/post-worker.js
ThaUnknown: This was more to reduce useless/duplicate code and make it more readable, but I'll define more functions for these

ThaUnknown: 375a943ae4324bed123d8c6db7fb1ec03f785c92

src/post-worker.js
ThaUnknown: 375a943ae4324bed123d8c6db7fb1ec03f785c92
does this fix it?

TFSThiagoBR98: Yes, but my IDE conflict this variable with the window.screen object

ThaUnknown: > Yes, this fixes the build, but now my IDE conflict this variable with the window.screen object

this was defined previously too, and I can see it being used in emscripten, so I'm not touching it

starkana:
Need a cheerleader

ThaUnknown:
additionally this PR is a fix to #46 [as the renderer won't request more subtitles under load], partial fix to #72 [perf optimisations], full fix to #97 [as onDemandRender doesn't use event listeners, and destory() removes listeners], partial fix to [1001432683](https://github.com/libass/JavascriptSubtitlesOctopus/issues/72#issuecomment-1001432683)[as you can use _demandRender, omitting JSO's render loop], and partial fix to #8 [only for ondemand render]

TheOneric:
The patch requirements have been explained before and serve to ensure long-term maintainability and reducing the chance of new bugs slipping in. Reducing them to "no broken commits" is leaving out several key points. Most notably this fails at not bundling logically unrelated changes and proper documentation. Some of the listed changes also appear to have API implications. The list in OP with pointers to individual code section strongly indicates, that a separation is very much possible.

ThaUnknown:
> The patch requirements have been explained before and serve to ensure long-term maintainability and reducing the chance of new bugs slipping in. Reducing them to "no broken commits" is leaving out several key points. Most notably this fails at not bundling logically unrelated changes and proper documentation. Some of the listed changes also appear to have API implications. The list in OP with pointers to individual code section strongly indicates, that a separation is very much possible.

while separation is possible, this means loosing functionality in the initial PRs as i didn't modify JSO, but deleted all of it and started from 0, for my own sanity

are you okay with PRs like that?

TheOneric:
There must be “no temporary user-visible behaviour changes” *(and permanent changes need to be justified)*. Start from the existing code and change it one step to be review- and understandable.

ThaUnknown:
> There must be “no temporary user-visible behaviour changes” _(and permanent changes need to be justified)_. Start from the existing code and change it one step to be review- and understandable.

I don't see how that can happen with a full rewrite of the JS part of the library

to put it simply the current js is close to unmaintainable, be because it was made from a fork of a fork of a library, you even have some of the original libassjs code in here, taking the code apart like that to make massive changes will take weeks

Issue #103, opened 2021-06-19
Future fontconfig versions ⪈2.13.94 won't need the fcstat patch


vadimkantorov:
I found they have an accessible bug tracker at https://gitlab.freedesktop.org/fontconfig/fontconfig/-/issues and was thinking to report it there

https://github.com/libass/JavascriptSubtitlesOctopus/blob/master/build/patches/fontconfig/0003-fix-fcstats-emscripten.patch

I'm using your patch for compiling TexLive under emscripten: https://github.com/busytex/busytex/blob/main/fontconfig_emcc.patch . Thanks for your patch!

TheOneric:
Thanks @vadimkantorov for looking into this and contacting Fontconfig! As it turned out though, the version of emscripten we previously used, had a linker bug which made it not work even after the fix at Fontconfig’s side, but this should now be sorted out after upgrading emscripten.

PR #129, opened 2022-02-21
Split Makefile, Add build flags

TFSThiagoBR98:
This is some work for project cleanup

- Split the Makefile, add support for build flags (DEBUG, ENABLE_PTHREADS, ENABLE_LTO);
  - e.g.: `make ENABLE_LTO=true DEBUG=true`
- Migrate from autotools to meson, to reduce patches;
- Add mixed build type for both wasm and legacy support;
- Add .editorconfig;


.gitignore
```diff
@@ -32,3 +32,6 @@ node_modules/
 build/lib
 src/SubOctpInterface.cpp
 src/SubOctpInterface.js
+build_dir/
+src/build_dir/
```
eli-schwartz, line 36: Any of this that is used for Meson's working build directory are not needed -- Meson already creates a `build_dir/.gitignore` file containing `*` for you, on the principle that it is always correct to ignore an out of tree build directory, and in most cases the project doesn't know which directory the user selects (but of course Meson does know).

TFSThiagoBR98: Thanks, I didn't know that

eli-schwartz: I am one of the Meson developers. :)

It's a fun easter egg to tell people about.

Makefile
```diff
@@ -274,71 +58,181 @@ OCTP_DEPS = \
...
+	touch $(BASE_DIR)/build/lib/created
+
+include $(BASE_DIR)/build/m4/fribidi.mk
+include $(BASE_DIR)/build/m4/expat.mk
+include $(BASE_DIR)/build/m4/brotli.mk
+include $(BASE_DIR)/build/m4/freetype_bootstrap.mk
+include $(BASE_DIR)/build/m4/harfbuzz.mk
+include $(BASE_DIR)/build/m4/freetype.mk
+include $(BASE_DIR)/build/m4/fontconfig.mk
+include $(BASE_DIR)/build/m4/libass.mk
```
eli-schwartz, line 82: This will hopefully become largely obsolete at some point due to https://github.com/libass/libass/pull/330 -- then Meson can just do a mega-build of all the projects at once via its internal subproject support.

TheOneric:
I merged most of the small commits unrelated to the build system changes.
As for the git clean change, not removing untracked files from our root repo makes a lot of sense to me, but the benefit isn't as clear to me for the submodules. Do you have an actual use for untracked files in the submodules? Stray files there may taint the `licensecheck` results and cause issues on commit updates if they collide with newly added files.


I'll write more regarding the build system changes at a later date, but I can already see that this commit does too many unrelated things at once.

TheOneric:
> I'll split this into multiple PR, for better review the changes.

Multiple *commits* in one PR is probably sufficient. If you prefer, feel free to use multiple PRs though.

build/m4/brotli.mk
```diff
@@ -37,4 +39,7 @@ git-brotli:
 	cd lib/brotli && \
 	git reset --hard && \
 	git clean -dfx && \
-	git pull origin master
+	git fetch origin && \
+	git checkout $(BROTLI_GIT_VERSION) && \
+	git submodule sync --recursive && \
+	git submodule update --init --recursive
```
TheOneric, line 45: I dislike how this requires duplicating the submodule version information which is already stored in our root git itself.
Instead the following can do the same without duplication:
```sh
    cd lib/brotli && \
    git reset --hard && \
    git clean -dfx
    git submodule update --force lib/
```

On another note; those targets seem to have originally been intended to *update* our submodules all at once as the `git-update` name of the target invoking all other `git-*` targets suggests. This commit changes their purpose to resetting/syncing the working state to the HEAD state. Imho this is new purpose is more useful, but perhaps we should then also rename the `git-update` target to something like `git-smreset` or so?

Makefile
```diff
@@ -106,7 +106,7 @@ clean-dist:
 clean-libs:
 	rm -frv dist/libraries build/lib
 clean-octopus:
-	cd src && git clean -fdx
+	cd src && git clean -fdX
 
 git-checkout:
 	git submodule sync --recursive && \
```
TheOneric, line 112: Feel free to merge this one commit.

TheOneric (commented):
First regarding only the current 3 commits:

 - The `m4` subfolder is commonly used for M4 files containing macros used in eg autoconf's `configure.ac`. Putting Makefile fragments there does not seem intuitive to me.
- I once also  considered splitting the Makefile up similar to how you did now, but never got around to taking a shot at it. However looking at the effects now, I'm no longer so sure that this is actually helpful.
It hides even the targets other fragments depend on and there remains a lot of copy-and-paste which if the instacnes are scattered over multiple files I'm afraid are only even more likely to unintentionally drift apart over time introducing bugs.

SInce we are using GNU Make, an alternative approach could be defining callable macros for all common tasks it makes sense — possibly with the macro definitions moved into a different file. This would keep all the targets in one file, reduce copy-and-paste shortening the Makefile and likely reducing the surface for bugs.


All the `git-*` rules for example could be shortened into this:
*(Btw, we  currently do not mark meta-targets as such via `.PHONY` but should probably do so)*
```make
git-checkout:
    git submodule sync --recursive && \
    git submodule update --init --recursive

git-smreset: git-brotli git-expat git-fontconfig git-freetype git-fribidi git-harfbuzz git-libass

.PHONY: git-checkout git-smreset

define TR_GIT_SM_RESET
git-$(1):
    cd lib/$(1) && \
    git reset --hard && \
    git clean -dfx
    git submodule update --force lib/$(1)

.PHONY: git-$(1)
endef

$(foreach subm, brotli expat fontconfig freetype fribidi harfbuzz libass, $(eval $(call TR_GIT_SM_RESET,$(subm))))
```

Note that since it creates new Make-targets the expansion needs to be processed by Make by calling `eval`, to make it easier to know the macro (whose definition might be in a different file in a final version)* does, I tenatively used an `TR_` prefix for “creates a target”.

For configuration + build + install of projects using autoconf-like configure + make, we could use something like the below, shown only at the example of libass:
```make
define FN_COMPILE_GNU
    cd build/lib/$(1) && \
    EM_PKG_CONFIG_PATH=$(DIST_DIR)/lib/pkgconfig \
    emconfigure $(or $(CONFIGURE_PREFIX_$(1)),.)/configure \
        CFLAGS=" \
        -s USE_PTHREADS=0 \
        $(GLOBAL_CFLAGS) \
        -s NO_EXIT_RUNTIME=1 \
        -s MODULARIZE=1 \
        " \
        --prefix="$(DIST_DIR)" \
        --host=x86-none-linux \
        --build=x86_64 \
        --disable-shared \
        --enable-static \
        $(CONFIGURE_FLAGS_$(1)) \
    && \
    emmake make -j8 && \
    emmake make install
endef

CONFIGURE_PREFIX_libass := ../../../lib/libass
CONFIGURE_FLAGS_libass  := --disable-asm --enable-harfbuzz --enable-fontconfig
$(DIST_DIR)/lib/libass.a: $(DIST_DIR)/lib/libfontconfig.a $(DIST_DIR)/lib/libharfbuzz.a $(DIST_DIR)/lib/libexpat.a $(D
IST_DIR)/lib/libfribidi.a $(DIST_DIR)/lib/libfreetype.a $(DIST_DIR)/lib/libbrotlidec.a build/lib/libass/configured
    $(call FN_COMPILE_GNU,libass)
```

Note, that unneeded `CONFIGURE_…` variables do not need to be defined and the exapnded call does not need to be `eavl`ed.
Also we could probably use Make’s `export` directive for `EM_PKG_CONFIG_PATH` to also not need to put it into all the macros of the different build systems. Possibly we could also `export` the `GLOBAL_CFLAGS` as just `CFLAGS` and expand instead of overwrite the value when we want to do so. For further shortening we could also introduce a common `LIBDIR` or just `LIBD` variable for `$(DIST_DIR)/lib`.

What are your thoughts on this?

Makefile
TheOneric: There are also CMake-only projects currently in there and in the future more autotools-only projects might be added.
Can a “Meson mega-build” properly parallelise CMake and “Unstable External Project” subprojects. I'd guess the former could in theory work since CMake has a Ninja-backend, but with Ninja’s lack of jobserver support I do not see how the Make-using latter could be fully parallelised together with the rest of the build.
That is assuming the warnings in Meon's docs regarding mixing build systems and the scary “experimental”/“unstable” label of the latter do not cause issues.

TheOneric:
Next, some remarks from my notes on the initial version and not contained in the current commits:


 - Please do not require `bash` specifically if nothing specific to bash is being used. And iinm we currently do not use any bashisms in our build, so unless there’s a good reason we should probably continue to work with any POSIX shell.

- The old commit message the primary goal of the switch to Meson was to “reduce the number of patches.”. It doesn't appear those two are correlated at all:
   - The `harfbuzz/00002` patch seemed superfluous regardless of which build system is used
   -  According to #103 the Fontconfig fcstat patch is supposedly unneeded only after we bump our vendored Fontconfig version. If I didn't miss something, you did not bump the version but were removing the patch.
    I did test bumping fontconfig and removing the patch a couple of weeks ago: while it built, at runtime I was getting errors regarding fcstat, but hadn't yet have time to investigate further. Have you checked for runtime errors after you removed this patch?
   - Other than the two patches above it appears that the only patch which can be removed maybe because of meson is `freetype/0001-disable-exports`. In my testing removing the patch without changing to Meson, did not cause any build problems. What issues did this try to fix?

TheOneric:
Now regarding the switch the benefits and drawbacks of switching many subprojects to Meson in the context of JSO.

There are some time-savings in Meson using Ninja as a backend and that it compared to autotools from git, only does one configuration step and likely does it quicker. However, I suspect the bulk of the speed improvement you measured comes from the hardcoded eight jobs being replaced by ninja’s default behaviour of using as many jobs as there are logical CPUs available.
Eight jobs are too much for GitHub Action runners, which advertise two logical CPUs, and in local builds it may be either too much or too little. And in some cases, like FriBidi, our build always only uses one(!) thread *(probably the `-j 8` parameter was forgotten, something which could be avoided by using macros to avoid copy-paste)*.
This by itself could also be easily achieved with Make, but sure using Meson also gives the other benefits listed before, so why not.

However, regarding speed there’s one thing that would be even better than using a more appropiate job count for each project separately and which is possible with Make but not Ninja (and therefore Meson if it cannot do a “mega-build” in which all projects use Meson): parallelise not only the individual (sub)project builds but also across all targets by using a jobserver. This means if one project no longer needs all the threads, another target can already be executed.
I started an attempt at this a few months ago, but it seemed like (probably) the `emmake` wrapper interfered with the relevant jobserver arguments. I had the idea of removing all the `emmake`s in the Makefile and instead calling the root Makefile with `emmake`, but never actually did because other things came up which took precedence.
***If*** the jobserver can be made to play nicely with `emmake` the gains from it may possibly outperform the Ninja and configure related gains from Meson. The catch is HarfBuzz having deprecated its autotools build, so once we switch HB to Meson, it either will consume more threads than we’d like or we need to ensure HarfBuzz specifically is not run in parallel with other builds.

There also is a patch series (ninja-build/ninja/#1140) to add jobserver support to Ninja since almost 6 years which is still maintained and updated. The last response from a Ninja dev is from 2 years ago though, which doesn't exactly inspire confidence in this getting merged anytime soon.

Since we have multiple CMake-only, at least one autotools-only with prospective two more additional projects, a Meson-and-ninja-only “mega-build” seems unlikely to happen. Using top-level meson with experimental non-Meson subprojects will also suffer from not being able to parallelise across projects *(in addition to it being experimental and the warnings from the docs)*.

Makefile
eli-schwartz: In the case of cmake, Meson actually runs cmake in trace mode and extracts the information about targets, then converts them to Meson rules and builds everything directly in Meson's own build.ninja regardless.

In the Unstable External Project case, Meson has no way of knowing what make targets need to be built. So it just builds everything, and updates a `name.stamp` file on completion. Any individual target that depends on an autotools external project will have that target do a complete build of the autotools project.

Any target that does NOT depend on the external project can be parallelized with the invoked `make` command, but you are indeed correct that ninja doesn't have any jobserver support so this isn't quite so helpful.

In fact, it may even be worse. :( the invoked `make` gets `-j<nproc>` so make and ninja actually compete for jobs, and `ninja -j1` isn't respected by make at all (and it cannot even try! There's no way to probe for that info from a sub-job of ninja. Make would actually set $MAKEFLAGS based on your arguments.)

The good news is that fribidi, freetype, fontconfig and harfbuzz have meson.build support upstream, while expat and brotli (which both officially support cmake as well as autotools) both have [WrapDB](https://wrapdb.mesonbuild.com/) support via third-party maintained meson.build files hosted through Meson's official index.

So you would never have needed to touch the Unstable External Project module, only cmake, and you don't really need to do that either due to the WrapDB versions.

eli-schwartz:
Which are the cmake-only projects?

freetypeand harfbuzz both support Meson, cmake, and configure+make all at once, very dedicated of them.
fontconfig and fribidi both support meson and configure+make

expat and brotli both support cmake and configure+make

libass is autotools only, but has work on porting to meson.

None of those seem to be cmake-only, and only libass itself is configure+make only.

Makefile
TheOneric: > while expat and brotli (which both officially support cmake as well as autotools) both have [WrapDB](https://wrapdb.mesonbuild.com/) support via third-party maintained meson.build files hosted through Meson's official index.

The linked site indicates that the current expat wrap is version `2.29-4` *(I'm assuming this means upstream version 2.2.9, 4th iteration of the wrap for the identical upstream release)* and brotli is at `1.0.7-1`. Those upstream versions were released 2019-09-25 and 2018-10-23 respectively which is older than what JSO currently uses and with several new upstream releases since fixing security bugs.
Also JSO needs to patch some projects and their build-system to disable certain things which do not play nicely with the WASM runtime in common browser configurations. Relying on both a third-party build-system which at most updates on new releases and then our own patches at top seems more fragile than necessary.

For the above reasons, I think relying on WrapDB is not the way to go for JSO.

TheOneric:
Btw, since JSO compiles from git not from a tarball, the difference between a proper standard autotools setup (e.g. libass) and custom config+make systems (in FreeType’s case a thin but required wrapper around autools) is relevant as the configuration steps differ.

> expat and brotli both support cmake and configure+make

I missed them supporting an autotools build; then it’s 2× CMake-or-autotools not CMake-only. Doesn't make much of a difference though, since both have a Make backend and are not Meson.
WrapDB doesn't seem suitable as explained in another comment.

> libass is autotools only, but has work on porting to meson.

That port only aims for Meson as a second-class system missing some features and e.g. X32, possibly enough for JSO though. Also not certain when this comes to a conclusion. But then still, libunibreak and libiconv also use Make-based systems and might be added in the future.

If I understood your explanation correctly, with a Meson-at-toplevel build using the experimental “external project” feature, the individual external projects and those controlled by Meson will use distinct thread pools which together can exceed the user-chosen limit. If so, using it does not seem desirable.
Our options then would be about if we want cross-(sub)project parallelisation. If yes, we should avoid ninja-backed systems when possible *(and take a look if HarfBuzz’s CMake system will continued to be supported or is also deprecated)*. If not, using Meson when provided as a first-class build-system is likely slightly faster than using a Make-based system. In both cases we would stick to GNU Make at top-level.
If at some point in the future Ninja gains jobserver-client support, we can with GNU Make at the toplevel still get both Meson for the subprojects *and* cross-project parallelisation.

Makefile
eli-schwartz: > Relying on both a third-party build-system which at most updates on new releases

Build system fixes are accepted at any time, updating to unstable snapshots is not.

These are updated as time allows and when people are interested in working on them -- typical OSS. I just looked at brotli, it was totally trivial to update the version and update it to modern `[provide]` standards on the side. :)

I could take a look at expat soon, probably.

TheOneric: I appreciate your work on updating them now and I'm sure the users of those wraps will do too. But in JSO, (a) we also will want to update them again sometime in the future and I'm not thrilled to maintain a bunch of full buildsystems for our dependencies and (b) JSO does not always use only releases of its subprojects. *(Currently all but libass is pointed at a release iinm, but historically and possibly in the future that won't/wasn't always the case. It definitely won't be the case for libass at least.)*

eli-schwartz:
> and take a look if HarfBuzz’s CMake system will continued to be supported or is also deprecated

Eh... it was removed once already, and then reverted and re-added with the rationale that "there is no need to rush" to remove it, to prevent people being caught off guard. They state in various PRs that they're not keeping it around forever.

> If at some point in the future Ninja gains jobserver-client support, we can with GNU Make at the toplevel still get both Meson for the subprojects _and_ cross-project parallelisation.

But you can use Kitware's ninja fork which does have jobserver-client support, but not jobserver support -- this can generally be worked around even for a ninja-first build with a simple Makefile:
```make
all:
	@+ninja
```

It is true that at the rate things are going, the ninja devs will likely never merge it -- but, Kitware regularly rebases that patchset for new releases and this is e.g. where `pip install ninja` is sourced from. There are people very invested in that feature. Just don't bother with the one from Debian's repository in that case.

Makefile
TheOneric: Merged it.

PR #132, opened 2022-02-28
fix resize

turuslan:
changes:
- fix "top/left changed" check.
  - was comparing `style.top`/`style.left` (string with `px`suffix) with number.
- split "top/left changed" and "width/height changed" code.
  - `left`/`top` and `width`/`height` are not related.

fixes:
- canvas became empty because of unnecessary resize.
  - open [videojs.html](https://libass.github.io/JavascriptSubtitlesOctopus/videojs.html).
  - pause when subtitles are visible.
  - resize window by 1 pixel, video/canvas size is not changed.
  - subtitles disappear.

### DIFF
--- a/src/subtitles-octopus.js
+++ b/src/subtitles-octopus.js
@@ -453,14 +453,9 @@ var SubtitlesOctopus = function (options) {
 
 
         if (
-          self.canvas.width != width ||
-          self.canvas.height != height ||
-          self.canvas.style.top != top ||
-          self.canvas.style.left != left
+          self.canvas.style.top != top + 'px' ||
+          self.canvas.style.left != left + 'px'
         ) {
-            self.canvas.width = width;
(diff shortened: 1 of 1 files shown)


dmitrylyzo:
Jellyfin's `Render Ahead` mode doesn't redraw the frame and is forced to resize the canvas according to the rendered viewport, so the browser scales the result itself.

So the fix works for non-cached draws (at least `wasm-blend`), but not for our mode.

[This](https://github.com/dmitrylyzo/JavascriptSubtitlesOctopus/commit/6f33693ff6c438395d9f098df13936de36dd3576) works for our mode (and `wasm-blend`) - 3840x2160 frame (canvas size) can be drawn on 1280x720 canvas (style size).
But I'm not sure about using styles for the canvas that isn't driven by the video element.

PR #133, opened 2022-03-24
Buffer refactor and buffer bug fixes

TheOneric:
The refactor into a class is part of the jellyfin backports and here followed up by various bug fixes. The class members have been refactored for clarity and unused functions dropped.
As I mentioned in [#120](https://github.com/libass/JavascriptSubtitlesOctopus/pull/120#issuecomment-1030927083), checking the size on each frame rather than only on resizes is somewhat superfluous now *(but already what happened before this change)* but apparently the "smart-blend" changes will require resizes on each frame. Let’s wait until those patches come in to see if this can be improved upon.

### DIFF
--- a/src/SubtitleOctopus.cpp
+++ b/src/SubtitleOctopus.cpp
@@ -43,11 +43,12 @@ public:
     /*
      * Request a raw pointer to a buffer being able to hold at least
      * x times y values of size member_size.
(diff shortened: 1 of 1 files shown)


TheOneric:
If anyone wants to take a closer look at this, write so now. If no one requests more time or objects, I'll merge this in about 2 days.

PR #134, opened 2022-04-02
Jellyfin’s dropAllAnimations

TheOneric:
Adds the corrected core animation detection and deletion code for `dropAllAnimations` and `renderAhead` and backports `dropAnimations` from jellyfin.

@dmitrylyzo Could you take a look at this to check if I missed to backport any relevant fixes for `dropAllAnimations`? In particular there’s https://github.com/libass/JavascriptSubtitlesOctopus/pull/111/commits/08f72cd094f7ee477a3dc8645101fb40d8f413a9 which refers to “lite mode” in its title, but iiuc the actual changes are all in the `renderAhead` portion.

### DIFF
--- a/src/SubtitleOctopus.cpp
+++ b/src/SubtitleOctopus.cpp
@@ -106,6 +106,131 @@ public:
     unsigned char* image;
 } RenderBlendResult;
 
+/**
+ * \brief Overwrite tag with whitespace to nullify its effect
+ * Boundaries are inclusive at both ends.
+ */
+static void _remove_tag(char *begin, char *end) {
+    if (end < begin)
+        return;
+    memset(begin, ' ', end - begin + 1);
+}
+
+/**
+ * \param begin point to the first character of the tag name (after backslash)
+ * \param end   last character that can be read; at least the name itself
+                and the following character if any must be included
+ * \return true if tag may cause animations, false if it will definitely not
+ */
+static bool _is_animated_tag(char *begin, char *end) {
+    if (end <= begin)
+        return false;
+
+    size_t length = end - begin + 1;
+
+    #define check_simple_tag(tag)  (sizeof(tag)-1 < length && !strncmp(begin, tag, sizeof(tag)-1))
+    #define check_complex_tag(tag) (check_simple_tag(tag) && (begin[sizeof(tag)-1] == '(' \
+                                        || begin[sizeof(tag)-1] == ' ' || begin[sizeof(tag)-1] == '\t'))
+    switch (begin[0]) {
+        case 'k': //-fallthrough
+        case 'K':
+            // Karaoke: k, kf, ko, K and kt ; no other valid ASS-tag starts with k/K
+            return true;
+        case 't':
+            // Animated transform: no other valid tag begins with t
+            // non-nested t-tags have to be complex tags even in single argument
+            // form, but nested t-tags (which act like independent t-tags) are allowed to be
+            // simple-tags without parentheses due to VSF-parsing quirk.
+            // Since all valid simple t-tags require the existence of a complex t-tag, we only check for complex tags
+            // to avoid false positives from invalid simple t-tags. This makes animation-dropping somewhat incorrect
+            // but as animation detection remains accurate, we consider this to be "good enough"
+            return check_complex_tag("t");
+        case 'm':
+            // Movement: complex tag; again no other valid tag begins with m
+            // but ensure it's complex just to be sure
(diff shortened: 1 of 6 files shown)


src/SubtitleOctopus.cpp
```diff
@@ -106,7 +106,130 @@ typedef struct RenderBlendResult {
...
+            case '{':
+                // Escaping the opening curly bracket to not start an override block is
+                // a VSFilter incompatiple libass extension. But we only use libass, so...
+                if (!block_start && (p == event->Text || *(p-1) != '\\'))
+                    block_start = p;
+                break;
+            case '}':
+                if (block_start && p - block_start > 2) {
+                    bool is_anim = _is_block_animated(block_start + 1, p - 1, drop_animations);
+                    if (is_anim && !drop_animations)
```
dmitrylyzo, line 214: _Visually, the suggestion has not been verified_
If `drop_animations: true`, `_is_block_animated` will never return `true` (all animations have been discarded).
```suggestion
                    if (_is_block_animated(block_start + 1, p - 1, drop_animations))
```

src/SubtitleOctopus.cpp
```diff
@@ -106,7 +106,130 @@ typedef struct RenderBlendResult {
...
+
+    // Search for override blocks
+    // Only closed {...}-blocks are parsed by VSFilters and libass
+    char *block_start = NULL; // points to opening {
+    for (char *p = event->Text; *p != '\0'; p++) {
+        switch (*p) {
+            case '{':
+                // Escaping the opening curly bracket to not start an override block is
+                // a VSFilter incompatiple libass extension. But we only use libass, so...
+                if (!block_start && (p == event->Text || *(p-1) != '\\'))
```
dmitrylyzo, line 208: Shouldn't the escaping be parsed with the boolean flag?
For something like `bla-bla \\{\pos(100,200)}`.
_libass doesn't seem to escape the backslash either: `1\\N2` gets a new line._

src/SubtitleOctopus.cpp
```diff
@@ -124,12 +247,34 @@ class SubtitleOctopus {
...
+        drop_animations = !!value;
+        if (drop_animations)
+            scanAnimations(scanned_events);
+    }
+
+    /*
+     * \brief Scan events starting at index i for animations
+     * and discard animated tags when found.
+     * Note that once animated tags were dropped they cannot be restored.
+     * Updates the class memeber scanned_events to last scanned index.
```
dmitrylyzo, line 268:
```suggestion
     * Updates the class member scanned_events to last scanned index.
```

src/SubtitleOctopus.cpp
```diff
@@ -124,12 +247,34 @@ class SubtitleOctopus {
...
+     */
+    void scanAnimations(int i) {
+        ASS_Event *cur = track->events;
```
dmitrylyzo, line 271: unused?

dmitrylyzo (commented):
I have not tested this in action. Only the main part in that big PR, and it seemed to work fine. The rest - just visually.

src/SubtitleOctopus.cpp
TheOneric: Fixed, thx!

src/SubtitleOctopus.cpp
TheOneric: Appears to have been a leftover from a previous version. Now removed, thanks.

src/SubtitleOctopus.cpp
TheOneric: Sorry, I don't understand what your suggesting with a “boolean flag”. Can you elaborate?
Also there aren't any backslashes being escaped being checked, but backslashes escaping curly brackets (working only in libass).

src/SubtitleOctopus.cpp
TheOneric: Yup, you're right. Dropped the superfluous `!drop_animations` check and merghed the `if`s. Thanks!

src/SubtitleOctopus.cpp
dmitrylyzo: I meant that the condition only checks the previous backslash, but there can be another one before it. In this case something like `if (*p == '\\') escaped = !escaped;` is required. But since libass (SSA/ASS standard) has no escaping logic, the current implementation is fine.

TheOneric: Great, then it seems this is ready to be merged.

PR #135, opened 2022-04-02
Patch cleanup

TheOneric:
This reduces our library patches to only the pthread-disabling ones.

For the fcstat fontconfig patch to be dropped updating emscripten was also necessary due to a bug in the currently used version. Hoping to avoid emscripten-related changes taking over, I stuck to the last 2.x release instead of updating to 3.x, but even so some cleanup was required.

There are some points I’d like to hear an opinion on before proceeding:
 - If we want to drop the deprecated API instead of surpressing the warnings, I’ll just drop it from this PR and defer dropping to a future patch.
 - Since we only build static libs, the only linking happening (currently) is at the final step. Instead of GLOBAL_LDFLAGS we could also specify LDFLAGS as part of  only at the final linking step instead of setting them globally. Not sure what’s better.
 - If (not yet tested) the filesystem gets linked automatically, should we completly remove its flag entirely or keep it for verbosity?

Furthermore, the update introduced one more warning, but I wasn’t sure what to do about it — though everything still seems to work fine.
```
/emsdk/upstream/emscripten/em++ -Wall -Wno-deprecated -g -O2 -r -s ENVIRONMENT=web,webview,worker -s NO_EXIT_RUNTIME=1 -s MODULARIZE=1 -o subtitlesoctopus.bc SubtitleOctopus.o
em++: warning: generating an executable with an object extension (.bc).  If you meant to build an object file please use `-c, `-r`, or `-shared` [-Wemcc]
```
However, as one can see `-r` already *is* used and was even added explicitly?

---

Regarding the HarfBuzz "dependency" patch: in the first version of #129 the patch seems to have become unnecessary because both the bootstrap and final FreeType libs were installed into the same DESTDIR, so the old pcfile became automatically available. However, this also means the library files already existed albeit without HarfBuzz support. While this might have worked in a sequential build, this seems prone to race-conditions on full-parallel builds so I'm not in favour of adopting this approach.
As for why just dropping the patch without any changes intially seemed to work for me: I didn't `clean` before the test so the old harfbuzz-supporting FreeType pc-file was still present in the path. It cannot be dropped without an replacement in a clean build.


TheOneric:
Regarding the `-r` warning: it even appears when calling `em++ -r -o *.bc *.cpp` standalone without a compile only step first. Using just the compile `-c` without any pseudo-linking step solves the issue. With the autotools setup, the better approach for this would have been to tell automake to build a static library instead of declaring to build an executable program with modified linking parameters.
However, most of the configure test are currently ignored and we only target one toolchain anyway, so I replaced the `src/` build setup with just one custom Makefile which is shorter than both the current setup and the [Meson setup from the initial](https://github.com/libass/JavascriptSubtitlesOctopus/commit/196d76545376e1ebf449646da22e75c6ada894a0#diff-f28598af2e23aa5d2bc7c72e022ae2c56a33802eb970afffaeca1e40607f97fe=) #129 and also allows us to drop the configuration targets.

TheOneric:
Rebased and made a small change regarding the `MODULARIZE` flag; it is now removed instead of added to global flags.

If you plan to review this and need more time, say so soon and this can wait however long required. Otherwise I'll assume there are no objections and will merge this soon.

The now removed RFC messages in the commits were:
 - **src/make: suppress deprecation warnings**:
    Or should we just drop the deprecated API?
 - **make: split global linker and compile flags**:
   Alternatively, do not set LDFLAGS globally only at the final step as this is the only instance of linking (currently at least).
 - **make: remove filesystem flags from dependency builds**:
    modern emscripten is supposedly able to detect this automatically. If it works, should we also remove it from the final link flags?

PR #144, opened 2022-06-21
Ignore all files for Docker build context

WeebDataHoarder:
Every time a build without `$FAST`, `$ docker build` is invoked.
While the Dockerfile does not require any files from the local context, the current directory is loaded as part of the build process.

Before:
```
$ ./run-docker-build.sh
Sending build context to Docker daemon  604.9MB
...
```

After:
```
$ ./run-docker-build.sh
Sending build context to Docker daemon  3.072kB
...
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+# Ignore everything for build image step, /code is passed via a volume
+*


TheOneric:
Thanks!

Just a cosmetic nit: Could you move the `prevents ...` part into the message body, so the title doesn't overflow? And ideally also add a note to the body that buildah also parses the `.dockerignore` file (if no `.containerignore` is present).

WeebDataHoarder:
Extended the message body with the suggested changes

PR #146, opened 2022-06-22
Implement lazy font loading, enable embedded font extraction

WeebDataHoarder:
Contains one commit (7a69b7fcdee7b66f698ec032d4bc42abe86d85fe) from #145.

These changes extract the embedded font, and add the option to enable lazy font loading (where font resources are mapped, but not downloaded until libass tries to use them).

Additionally, `fallbackFont` option can be used to override the default fallback font that gets used. This way, users can specify their preferred font, including CJK ones (for example Noto Sans CJK .ttc).

Enabling lazy font loading is done via `lazyFontLoading` set to `true`, but it defaults to `false`.
This is necessary given some requirements from WASM to not fetch the whole file, specifically:
* HEAD and Range request support
* `Access-Control-Expose-Headers: Accept-Ranges, Content-Length, Content-Encoding` (at least)
* No Content-Encoding set.
If these conditions match, HEAD request will be issued and file will be accessed via Ranges as-needed. Otherwise content will be fetched as a whole file.

Additionally, this PR enables embedded font extraction support on libass side. This is split on a separate commit.

### DIFF
--- a/src/SubtitleOctopus.cpp
+++ b/src/SubtitleOctopus.cpp
@@ -6,6 +6,7 @@
 #include <stdlib.h>
 #include <stdarg.h>
 #include <string.h>
+#include <string>
 #include <cstdint>
 #include <ass/ass.h>
 
@@ -247,6 +248,8 @@ public:
 
     int status;
 
+    std::string defaultFont;
+
     SubtitleOctopus() {
         status = 0;
         ass_library = NULL;
@@ -281,7 +284,11 @@ public:
         scanned_events = i;
     }
 
-    void initLibrary(int frame_w, int frame_h) {
+    void initLibrary(int frame_w, int frame_h, char* default_font) {
+        if (default_font != NULL) {
+            defaultFont.assign(default_font);
+        }
+
         ass_library = ass_library_init();
         if (!ass_library) {
             fprintf(stderr, "jso: ass_library_init failed!\n");
@@ -353,11 +360,11 @@ public:
     void reloadLibrary() {
         quitLibrary();
 
-        initLibrary(canvas_w, canvas_h);
+        initLibrary(canvas_w, canvas_h, NULL);
     }
 
     void reloadFonts() {
-        ass_set_fonts(ass_renderer, "/assets/default.woff2", NULL, ASS_FONTPROVIDER_FONTCONFIG, "/assets/fonts.conf", 1);
+        ass_set_fonts(ass_renderer, defaultFont.c_str(), NULL, ASS_FONTPROVIDER_FONTCONFIG, "/assets/fonts.conf", 1);
     }
 
     void setMargin(int top, int bottom, int left, int right) {
--- a/Makefile
+++ b/Makefile
@@ -156,16 +156,16 @@ EMCC_COMMON_ARGS = \
 	$(LDFLAGS) \
 	-s AUTO_NATIVE_LIBRARIES=0 \
 	-s EXPORTED_FUNCTIONS="['_main', '_malloc']" \
+	-s DEFAULT_LIBRARY_FUNCS_TO_INCLUDE="['\$$Browser']" \
 	-s EXPORTED_RUNTIME_METHODS="['ccall', 'cwrap', 'getValue', 'FS_createPreloadedFile', 'FS_createPath']" \
-	--use-preload-plugins \
-	--preload-file assets/default.woff2 \
-	--preload-file assets/fonts.conf \
+	--embed-file assets/fonts.conf \
 	-s ALLOW_MEMORY_GROWTH=1 \
 	-s NO_FILESYSTEM=0 \
+	--memory-init-file=0 \
 	--no-heap-copy \
 	-o $@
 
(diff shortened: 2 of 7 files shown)


ThaUnknown:
is removing --no-heap-copy a good idea? https://github.com/emscripten-core/emscripten/issues/5179

WeebDataHoarder:
That is a rebase conflict, will force-push

src/SubtitleOctopus.cpp
```diff
@@ -247,6 +247,8 @@ class SubtitleOctopus {
 
     int status;
 
+    char defaultFont[256];
```
rcombs, line 250: Any particular reason for the fixed-sized buffer? Can we not use `std::string` here, or `std::unique_ptr`, or at worst, just `strdup` and an explicit `free` in the destructor?

shoghicp: No particular reason, this is how it was done originally for simplicity.
I'll change the code after rearranging commits from #145

WeebDataHoarder: No particular reason, this is how it was done originally for simplicity.
Will change this after rearranging #145 commits.

WeebDataHoarder: This is now using std::string

WeebDataHoarder:
Additional comment for `--embed-file`: since 3.1.3 (maybe an update could be worth it) it goes directly in memory, without efficiency loss. This is not an issue as mentioned on https://github.com/libass/JavascriptSubtitlesOctopus/pull/145#issuecomment-1163088861 given the few bytes that are being embedded.

WeebDataHoarder:
As far as I know it's only exposed in the worker, probably for debugging reasons. 

There are various spots where `ass_track` gets re-initialized in `post-worker.js`, setTrack() for example. `ass_renderer` currently does not get re-initialized. 

Most of the oct_add_font or other stuff, is only available to direct users of the worker.js (not jso module interface).

I can update the code so these values get updated properly, or rid of them elsewhere? Even if desired, you can always inspect/grab these handles via `self.octObj.ass_library` in worker code / debugger.

TheOneric:
My intent was that if JSO users can get the internal library the change to also clear memory fonts (and reinit all handles) needs to be documented. If no one has access to it, its an internal change and the existing high level description of "allows to change the track" or similar is sufficient for the affected API functions.

> There are various spots where ass_track gets re-initialized in post-worker.js, setTrack() for example.

I might have missed something, but grepping through the code it seems like those values are actually only getting assigned to and never used? *(If that’s the case: why do they even exist?)*

> Most of the oct_add_font or other stuff, is only available to direct users of the worker.js (not jso module interface).

Sorry, can you explain howsuch a direct use for a JSO user might look like, so that `oct_*` are available? I'm trying to understand if there is any more conflict potential and how the scopes here are or are not separated.

Both `oct_*` and `octObj` are mentioned in 4.0.0’s release notes and appear to have been added in #64; CC: @TFSThiagoBR98.

TheOneric:
Thanks, if there's no way an user can modify the library and renderer states or add memory fonts, then there’s nothing more to worry about regarding embedded fonts. Eventhough, the actually diff ended up quite concise, the interaction with the rest of JSO and the effects aren't, so how about a more verbose commit message similar to this? *(also avoiding overly long lines)*:
```
Use embedded fonts

ASS subtitles can embedd fonts with a custom encoding in
their [Fonts] section. For historic reasons ass_set_extract_fonts
defaults to disabled, so we need to explicitly opt-in.
There’s currently also an issue, with embedded fonts outliving
their track, which can lead to indefinitely growing memory consumption
if not all memory fonts are cleared on track reinit.

However, ass_clear_fonts can only be safely called if the renderer
also has been released first. At this point it is simpler to just
reinit the whole library-renderer-track triplet and library inits
are not that costly anyway.

JSO never uses (non-embedded) memory fonts itself and does not expose
any way to add them so this does not constitute a user-visible change.
Even oct_add_font cannot be used by consumers of upstream JSO binaries.

Note: the JS pointers of the libass handles are updated in this patch,
      but it appears they are actually never used.
```

If you can update the commit message and move the embedded font commit before the lazy loading one, I'm going to merge the embedded font commit in ~24h if rcombs or TFSThiagoBR98 don't object.

---

The question of how to use the interface and dead code warrants a dedicated thread indeed. Only if there would have been *some* way to add memory fonts there had been direct relevance to the change at hand. *(I just noticed `createTrackMem` also seems to be unused...)*

WeebDataHoarder:
I have reworded the embedded fonts commit, and moved it before lazy font loading.

---

On the topic of dead code, I have cleaned up a few of those and refactored a few methods to de-duplicate some code.
Might open a PR just for dead-code/feature elimination with those changes to start with.

TheOneric:
Merged embedded font commit.

src/SubtitleOctopus.cpp
```diff
@@ -5,7 +5,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <stdarg.h>
-#include <string.h>
+#include <string>
```
rcombs, line 8: `<string.h>` should still be pulled in here; we use e.g. `memcmp`. C++ `<string>` is separate and is not required to pull in `<string.h>`. If we want to use all C++ names, we could use `<cstring>`, but that's out of scope for this PR.

rcombs (commented):
Just one header change; otherwise LGTM.

rcombs (approved):

